Issue #5, opened 2019-04-21
Add notes to app


BintLopez:
- lockbox action has many notes

kyaroch:
Resolved in https://github.com/MidwestAccessCoalition/lockbox_rails/pull/17

PR #17, opened 2019-04-21
Create notes

kyaroch:
https://github.com/MidwestAccessCoalition/lockbox_rails/issues/5

### DIFF
--- a/db/migrate/20190421191047_create_notes.rb
+++ b/db/migrate/20190421191047_create_notes.rb
@@ -1,0 +1,13 @@
+class CreateNotes < ActiveRecord::Migration[6.0]
+  def change
+    create_table :notes do |t|
+      t.text :text
+      t.bigint :notable_id
+      t.string :notable_type
+
+      t.timestamps
+    end
+
+    add_index :notes, [:notable_type, :notable_id]
+  end
+end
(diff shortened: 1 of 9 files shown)


db/schema.rb
```diff
@@ -22,7 +22,9 @@
     t.bigint "lockbox_partner_id"
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
+    t.bigint "support_request_id"
```
kyaroch: I unintentionally pulled these changes in from https://github.com/MidwestAccessCoalition/lockbox_rails/pull/16. Let's just merge that one first, unless it needs changes, in which case I can fix this one

db/migrate/20190421191047_create_notes.rb
```diff
@@ -0,0 +1,13 @@
+class CreateNotes < ActiveRecord::Migration[6.0]
+  def change
+    create_table :notes do |t|
+      t.text :text
+      t.bigint :notable_id
```
kyaroch: This should probably `belong_to` a `User` as well, but we can add that once users are in master.

BintLopez:
Resolve conflicts first tho

PR #18, opened 2019-04-21
README and some basic setup

scottyschup:
* add better errors, binding of caller
* start creating README
* add note about using bootstrap to manifest

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -1,2 +1,5 @@
 //= link_tree ../images
 //= link_directory ../stylesheets .css
+
+// Add bootstrap
+// https://gorails.com/forum/install-bootstrap-with-webpack-with-rails-6-beta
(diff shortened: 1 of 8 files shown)


app/assets/config/manifest.js
```diff
@@ -1,2 +1,5 @@
 //= link_tree ../images
 //= link_directory ../stylesheets .css
+
+// Add bootstrap
+// https://gorails.com/forum/install-bootstrap-with-webpack-with-rails-6-beta
```
BintLopez: might not need this depending on how the USWDS stuff goes

scottyschup: Right. I mostly put this there because even if we don't use bootstrap, the pattern for setting up any CSS library via Webpack should be somewhat similar.

misternu:
jane mailcatcher copy:

#### Set up local email interceptor

```sh
gem install mailcatcher
mailcatcher
```

BintLopez:
Went ahead and merged the Readme part of this PR into master https://github.com/MidwestAccessCoalition/lockbox_rails/pull/49

PR #30, opened 2019-04-21
Creation of support requests

kyaroch:


app/models/lockbox_action.rb
```diff
@@ -94,4 +99,12 @@ def cancel!
   def complete!
     update!(status: COMPLETED)
   end
+
+  private
+
+  def inherit_lockbox_partner_id
+    if lockbox_partner_id.blank? && support_request&.lockbox_partner_id
+      self.lockbox_partner_id = support_request.lockbox_partner_id
```
BintLopez: good call here

BintLopez:
image

PR #32, opened 2019-04-21
Adds RSpec, test data gems, and sets up CI

BintLopez:
- RSpec
- FactoryBot
- Faker
- CircleCI


BintLopez:
Based on some cursory searching, I think the build failure is a flake so I think we should merge and test that hypothesis -- https://discuss.circleci.com/t/one-of-our-builds-has-randomly-started-failing-on-the-bundle-install-stage/17068

PR #43, opened 2019-04-22
Set up realistic fake data

BintLopez:
- Adds seed file of fake data including mac users, lockbox partners, users, and transactions
- Adds factories
- Adds missing associations and tests

Issue #42, opened 2019-04-22
Add money-rails gem


BintLopez:
In order to do proper math with money, we should use the money gem... this handles some of the complexity surrounding rounding and money (ewwwww)

https://github.com/RubyMoney/money-rails

Issue #11, opened 2019-04-21
Add balance method to lockbox partner


BintLopez:
Balance method on LockboxPartner represents the amount of $$ in the cashbox as of current date

PR #45, opened 2019-04-23
Balance calculation

BintLopez:
https://github.com/MidwestAccessCoalition/lockbox_rails/issues/11
https://github.com/MidwestAccessCoalition/lockbox_rails/issues/42

- Adds money gem
- Balance calculation for a partner's lockbox
- Amount method for lockbox action 

This was branched off of https://github.com/MidwestAccessCoalition/lockbox_rails/pull/43


app/models/lockbox_action.rb
```diff
@@ -3,5 +3,21 @@ class LockboxAction < ApplicationRecord
...
+    :reconcile,
+    :support_client
+  ].freeze
+
+  def amount
+    lockbox_transactions.map(&:amount).sum
+  end
+
+  def cancel!
+    update!(status: 'canceled')
```
kyaroch: We should probably be using constants to specify the valid statuses. Not critical right now.

If these bang methods raise an exception, will that get caught somewhere?

spec/models/lockbox_action_spec.rb
```diff
@@ -4,4 +4,55 @@
...
+      # makes sense is reconciliation where the amount
+      # expected vs the amount counted
+
+      it "return zero" do
+
+      end
+    end
+
+    xcontext 'when action is in canceled status' do
+      # Does it make sense to return zero no matter
```
kyaroch: Is there a reason anyone would want to know what the amount would have been, had it not been canceled?

I think returning zero would be fine for now; the info will still be there in the database in case that functionality is ever needed.

app/models/lockbox_partner.rb
```diff
@@ -1,2 +1,22 @@
...
+      when 'debit'
+        balance -= action.amount
+      end
+      balance
+    end
+  end
+
+  def relevant_transactions_for_balance(exclude_pending: false)
+    excluded_statuses = exclude_pending ? %w(canceled pending) : %w(canceled)
+    lockbox_action_ids = LockboxAction.where(lockbox_partner_id: self.id).where.not(status: excluded_statuses).pluck(:id)
```
kyaroch: I think we could clean this up a bit by creating an `:exclude_pending` scope on `LockboxAction` (since the logic relating to `lockbox_actions` columns/values ideally should be in that class).

If we did that, and then added a `has_many: lockbox_transactions, through: lockbox_actions` association, could we just do `lockbox_actions.exclude_pending.lockbox_transactions`? Would that work? I haven't tested it, but I think it would. Although it occurs to me we'd then have to either conditionally apply the scope or pass an argument to it, which I guess is also a little awkward.

spec/factories/users.rb
```diff
@@ -0,0 +1,8 @@
+FactoryBot.define do
+  factory :user do
+    name         { Faker::Name.full_name }
+    email        { Faker::Internet.email }
+    password     { 'superstrongpassword12345' }
```
kyaroch: 🥇

app/models/lockbox_partner.rb
BintLopez: ooooh good call

spec/models/lockbox_action_spec.rb
BintLopez: I don't think there's a reason for this. Will go the return 0 route

app/models/lockbox_action.rb
BintLopez: will add exception handling and yes.... will constantize

app/models/lockbox_partner.rb
BintLopez: I used the scope... but can't use...
```ruby
lockbox_actions.excluded_statuses(excluded_statuses).lockbox_transactions
```
... because the return from `lockbox_actions.excluded_statuses(excluded_statuses)` is an active record collection whereas the `lockbox_transactions` are associated with each individual lockbox action instance.

app/models/lockbox_action.rb
BintLopez: And.... actually not going to add exception handling. My thoughts are that in the beginning we're going to want errors to be loud so we can find and fix them

app/models/lockbox_partner.rb
kyaroch: I had a feeling there might be some sort of issue there, but I forgot what. (This is giving me flashbacks to the first production Rails application I ever built, in 2015, which had some similar issues)

PR #46, opened 2019-04-23
Lockbox action creation class and validations

BintLopez:
Adding logic for lockbox action and associated transaction creation


app/models/lockbox_action.rb
```diff
@@ -17,6 +17,50 @@ class LockboxAction < ApplicationRecord
...
   scope :excluding_statuses, -> (*statuses) { where.not(status: statuses) }
 
+  # action_type should correspond with ACTION_TYPES
+  def self.create_with_transactions(action_type, params)
+    if !ACTION_TYPES.include?(action_type)
+      raise ArgumentError.new("The only Lockbox actions supported are #{ACTION_TYPES.join(',')}")
+    else
+      ActiveRecord::Base.transaction do
+        lockbox_action = create!(
+          eff_date: params[:date],
```
kyaroch: Can we change this, throughout, to use `params[:eff_date]` instead? That's the field name I have in the form (since it's the column name).

app/models/lockbox_action.rb
```diff
@@ -17,6 +17,50 @@ class LockboxAction < ApplicationRecord
...
+          action_type: action_type,
+          status: PENDING
+        )
+
+        case action_type
+        when :add_cash
+          lockbox_action.lockbox_transactions.create!(
+            eff_date: params[:date],
```
kyaroch: Remind me why we're storing the `eff_date` in both tables? It looks like it's always going to be the same in both, if `LockboxTransaction`s are always created using this method (although I don't know if that's the case).

app/models/lockbox_action.rb
BintLopez: you got it!!!

app/models/lockbox_action.rb
BintLopez: So... hmmm yeah it's probably not entirely necessary... 

We need the eff_date to live on the transaction (more straightforward for the balance https://github.com/MidwestAccessCoalition/lockbox_rails/blob/master/app/models/lockbox_partner.rb#L20-L21) -- so for lockbox actions with a more complicated breakdown, the date will be duplicated.

spec/models/lockbox_action_spec.rb
```diff
@@ -2,9 +2,36 @@
 
 describe LockboxAction, type: :model do
   it { is_expected.to belong_to(:lockbox_partner) }
+  it { is_expected.to belong_to(:support_request).optional }
   it { is_expected.to have_many(:lockbox_transactions) }
   it { is_expected.to have_many(:notes) }
 
+  describe '.create_with_transactions' do
```
BintLopez: Actually need to write these tests!!!!

Issue #26, opened 2019-04-21
MAC user should be able to invite a clinic user


kyaroch:
As a MAC user, I want to be able to invite a clinic user associated with a clinic, so that they can manage the lockbox.

PR #62, opened 2019-05-12
Lockbox partners users

misternu:


misternu:
for #26

app/views/devise/passwords/edit.html.erb
```diff
@@ -18,7 +18,7 @@
   </div>
 
   <div class="actions">
-    <%= f.submit "Change my password" %>
+    <%= f.submit "#{resource.password.present? ? "Change" : "Set"} my password" %>
```
BintLopez: I think the double quotes here within the double quotes should be changed to single quotes

PR #75, opened 2019-05-19
Add transaction history for lockbox partner

madisoncarr:
Hey Nicole, 
Here is the stuff that we worked on a couple weeks back.

### DIFF
--- a/app/models/lockbox_partner.rb
+++ b/app/models/lockbox_partner.rb
@@ -20,4 +20,8 @@ class LockboxPartner < ApplicationRecord
     lockbox_action_ids = LockboxAction.excluding_statuses(excluded_statuses).pluck(:id)
     LockboxTransaction.where(lockbox_action_id: lockbox_action_ids)
   end
+
+  def historical_actions
+    @all_actions ||= lockbox_actions.order(:eff_date)
+  end
 end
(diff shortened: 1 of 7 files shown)


app/models/lockbox_partner.rb
```diff
@@ -20,4 +20,8 @@ def relevant_transactions_for_balance(exclude_pending: false)
     lockbox_action_ids = LockboxAction.excluding_statuses(excluded_statuses).pluck(:id)
     LockboxTransaction.where(lockbox_action_id: lockbox_action_ids)
   end
+
+  def historical_actions
+    @all_actions ||= lockbox_actions.order(:eff_date)
```
kyaroch: It looks like what we want here is to order them from newest to oldest, which would be `.order(eff_date: :desc)`

BintLopez:
Displaying the lockbox action history on the lockbox partner card and on its individual view...

image

image

BintLopez:
Also, very apparent from this test data that the balance calculation is wrong 🤦‍♀ 💸

BintLopez:
Follow up items
- [ ] work on the display for the transaction history (diff color for the support request line items, adding in the display text for the action types)
- [ ] fix the balance

PR #78, opened 2019-05-24
fix the balance

BintLopez:


app/controllers/support_requests_controller.rb
```diff
@@ -9,6 +9,8 @@ def new
   end
 
   def create
+    # TODO -- REFACTOR THIS
+    # LockboxAction.create_support_request(all_params)
```
BintLopez: Can clean up / simplify support request creation by pulling out into an interactor claass

PR #85, opened 2019-05-26
Account signup and management 82

### LINKED ISSUE #82: Account signup and management
- As a new user completing the invite work flow I should be able to add my name when signing up for an account
- As an existing app user, I should be able to edit my account settings including my name.
- The account settings page should match the app styling rather than the login styling.

kyaroch:
Resolves #82 
* Allow user to edit name on account management page
* Allow user to edit name when setting initial password
* Use application layout for account management page
* Fix CSS issue that caused devise forms to render incorrectly when additional fields were added (possibly without additional fields given a small enough screen)

Screenshot_20190526_145226
Screenshot_20190526_145252

### DIFF
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -12,9 +12,12 @@ class Users::PasswordsController < Devise::PasswordsController
   # end
 
   # GET /resource/password/edit?reset_password_token=abcdef
-  # def edit
-  #   super
-  # end
+  def edit
+    @resource = resource_class.find_by(
+      reset_password_token: params[:reset_password_token]
+    )
+    super
+  end
 
   # PUT /resource/password
   def update
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+class Users::PasswordsController < Devise::PasswordsController
+  # GET /resource/password/new
+  # def new
+  #   super
+  # end
+
+  # POST /resource/password
+  # def create
+  #   super
+  # end
+
+  # GET /resource/password/edit?reset_password_token=abcdef
+  # def edit
+  #   super
+  # end
+
+  # PUT /resource/password
+  def update
+    super do |resource|
+      if resource.errors.empty?
+        resource.update(update_password_params)
+      end
+    end
+  end
+
+  protected
+
+  # def after_resetting_password_path_for(resource)
+  #   super(resource)
+  # end
+
+  # The path used after sending reset password instructions
+  # def after_sending_reset_password_instructions_path_for(resource_name)
+  #   super(resource_name)
+  # end
+
+  def update_password_params
+    # Devise does not use these params to update the password itself, hence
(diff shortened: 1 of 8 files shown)


app/controllers/users/registrations_controller.rb
```diff
@@ -2,7 +2,9 @@
 
 class Users::RegistrationsController < Devise::RegistrationsController
   # before_action :configure_sign_up_params, only: [:create]
-  # before_action :configure_account_update_params, only: [:update]
+  before_action :configure_account_update_params, only: [:update]
+
+  layout 'application', only: :edit
```
BintLopez: 💯

BintLopez:
Weird behavior:

- As a new user, when I'm completing the sign up flow, if I give my name but have errors in my password, the name field on the sign up form disappears and the name I previously provided is not saved
- As a current user, when I'm editing my account information, if I don't provide a password to confirm my changes, I get a completely different error page that does not follow the app layout

app/controllers/users/passwords_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  # end
+
+  # POST /resource/password
+  # def create
+  #   super
+  # end
+
+  # GET /resource/password/edit?reset_password_token=abcdef
+  def edit
+    @resource = resource_class.find_by(
```
kyaroch: This change resolves the issue with the preset name not being populated on the form.

By default, devise does not actually attempt to load the user from the database until the password reset request is submitted. I'm a little bit leery of this change because it leaks the following pieces of information:
1. Whether the token is valid or not
2. The name of the user associated with the token (if it was preset)

I think this *probably* doesn't matter. You can discover (1) anyway by attempting to submit the form; as for (2), the token namespace is large enough that it should not be possible to bruteforce it. But we need to think carefully about this. @BintLopez, what do you think?

app/controllers/users/registrations_controller.rb
```diff
@@ -2,7 +2,9 @@
 
 class Users::RegistrationsController < Devise::RegistrationsController
   # before_action :configure_sign_up_params, only: [:create]
-  # before_action :configure_account_update_params, only: [:update]
+  before_action :configure_account_update_params, only: [:update]
+
+  layout 'application', only: [:edit, :update]
```
kyaroch: Adding `:update` fixes issue with incorrect layout on failed submission

app/views/devise/passwords/edit.html.erb
```diff
@@ -4,6 +4,14 @@
   <%= devise_error_messages! %>
   <%= f.hidden_field :reset_password_token %>
 
+  <% # Coercing to a boolean because #nonzero? does not return a boolean %>
+  <% unless !!@resouce&.sign_in_count&.nonzero? %>
```
kyaroch: Changing this conditional fixes the issue with the name field disappearing on failed submit, but will still hide the field on subsequent (user-requested) password resets

app/controllers/users/passwords_controller.rb
misternu: we should separately block bruteforce attempts at finding valid password reset tokens

Issue #66, opened 2019-05-12
Add role to user (mac & partner users)


BintLopez:
- Add role column to User table
- Two possible roles:  'mac' and 'partner'
- MAC users should see MAC dashboard upon login
- Partner users should see their specific Lockbox dashboard upon login

nsiwnf:
Update : 
Possible roles are 'admin' or 'partner'

PR #86, opened 2019-05-26
#66 Add role to user

nsiwnf:
#66

Added "role" to user
Adding dashboard view based on "admin" or "partner" User role

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -1,5 +1,9 @@
 class DashboardController < ApplicationController
   def index
-    @lockbox_partners = LockboxPartner.all
+    if current_user.lockbox_partner_id?
+      @lockbox_partner = LockboxPartner.find(current_user.lockbox_partner_id)
+    else
+      @lockbox_partners = LockboxPartner.all
+    end
   end
 end
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -1,7 +1,7 @@
 class DashboardController < ApplicationController
   def index
-    if current_user.lockbox_partner_id?
-      @lockbox_partner = LockboxPartner.find(current_user.lockbox_partner_id)
+    if current_user.partner?
+      @lockbox_partner = current_user.lockbox_partner
     else
       @lockbox_partners = LockboxPartner.all
     end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -10,6 +10,19 @@ class User < ApplicationRecord
 
   scope :confirmed, -> { where.not(confirmed_at: nil) }
 
+  ROLES = [
+    ADMIN  = 'admin',
+    PARTNER = 'partner'
+  ].freeze
+
+  def admin?
+    role == ADMIN
+  end
+
+  def partner?
+    role == PARTNER
+  end
+
   private
 
   def password_required?
--- a/app/views/dashboard/index.html.erb
+++ b/app/views/dashboard/index.html.erb
@@ -6,9 +6,7 @@
   <% if current_user.partner? %>
     <%= render partial: 'dashboard/lockbox_partner', object: @lockbox_partner, as: :lockbox_partner %>
   <% else %>
-  </div>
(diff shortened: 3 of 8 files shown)


app/controllers/dashboard_controller.rb
```diff
@@ -1,5 +1,9 @@
 class DashboardController < ApplicationController
   def index
-    @lockbox_partners = LockboxPartner.all
+    if current_user.lockbox_partner_id?
+      @lockbox_partner = LockboxPartner.find(current_user.lockbox_partner_id)
```
BintLopez: For this you can just do `@lockbox_partner = current_user.lockbox_partner` instead of calling find

app/controllers/dashboard_controller.rb
```diff
@@ -1,5 +1,9 @@
 class DashboardController < ApplicationController
   def index
-    @lockbox_partners = LockboxPartner.all
+    if current_user.lockbox_partner_id?
```
BintLopez: For this, shall we use `current_user.partner?`

BintLopez:
Now that we're adding in the different roles, we actually need to set those when doing new user set up. As a part of this, we should also modify the new user creation to add the lockbox_partner_id and role when inviting a partner to the app

app/models/user.rb
```diff
@@ -10,6 +10,19 @@ class User < ApplicationRecord
 
   scope :confirmed, -> { where.not(confirmed_at: nil) }
 
+  ROLES = [
+    ADMIN  = 'admin',
+    PARTNER = 'partner'
+  ].freeze
```
misternu: This results in `['admin', 'partner']` did you mean for it to be a hash?

app/views/dashboard/index.html.erb
```diff
@@ -1,13 +1,15 @@
...
-  <div id="dashboard-lockbox-partners">
-    <%= render partial: 'dashboard/lockbox_partner', collection: @lockbox_partners %>
+  <h2 class="text-info text-center">
+    Welcome,
+    <%= current_user.name %>!</h2>
+
+  <% if current_user.partner? %>
+    <%= render partial: 'dashboard/lockbox_partner', object: @lockbox_partner, as: :lockbox_partner %>
+  <% else %>
   </div>
```
BintLopez: is this an extra </div> ?

BintLopez: Ahhhh... I think this should be the end for the div that's opened on line 11

app/views/dashboard/index.html.erb
```diff
@@ -1,13 +1,15 @@
...
-  <h4>Lockbox Partners (<%= @lockbox_partners.length %>)</h4>
-  <div id="dashboard-lockbox-partners">
-    <%= render partial: 'dashboard/lockbox_partner', collection: @lockbox_partners %>
+  <h2 class="text-info text-center">
+    Welcome,
+    <%= current_user.name %>!</h2>
+
+  <% if current_user.partner? %>
+    <%= render partial: 'dashboard/lockbox_partner', object: @lockbox_partner, as: :lockbox_partner %>
+  <% else %>
```
BintLopez: can we indent everything under the `else` ?

db/seeds.rb
```diff
@@ -19,7 +20,8 @@
   User.where(email: partner_user_email).first_or_create!(
     lockbox_partner: lockbox_partner,
     password: 'heytherefancypants4321',
-    confirmed_at: Time.current
+    confirmed_at: Time.current,
+    role: 'partner'
```
BintLopez:
```suggestion
    role: User::PARTNER
```

db/seeds.rb
```diff
@@ -1,6 +1,7 @@
 mac_user = User.where(email: 'cats@test.com').first_or_create!(
   password: 'password1234',
-  confirmed_at: Time.current
+  confirmed_at: Time.current,
+  role: 'admin'
```
BintLopez:
```suggestion
  role: User::ADMIN
```

app/views/dashboard/index.html.erb
nsiwnf: @BintLopez There are 2 div blocks here - should this just be one?

Issue #91, opened 2019-06-02
Add cash to the lockbox


BintLopez:
As an admin user, I can complete a web form to add an amount to a given lockbox.

- completing this form should create a lockbox action and transaction
- this should trigger an email to the lockbox partners users on that lockbox notifying them that more $$ is being sent their way

PR #98, opened 2019-06-09
Add cash to lockbox

kyaroch:
Backend implementation for https://github.com/MidwestAccessCoalition/lockbox_rails/issues/91


lib/add_cash_to_lockbox.rb
```diff
@@ -0,0 +1,46 @@
...
+        amount_cents: amount_cents,
+        balance_effect: LockboxTransaction::CREDIT
+      )
+
+      unless lockbox_transaction.valid?
+        err_message = "Lockbox transaction not created: #{lockbox_transaction.errors.full_messages.join(', ')}"
+        raise ActiveRecord::Rollback
+      end
+
+      lockbox_action
```
BintLopez: The call method should return the lockbox_action -- let's add a test for this

app/models/lockbox_transaction.rb
```diff
@@ -4,8 +4,8 @@ class LockboxTransaction < ApplicationRecord
   belongs_to :lockbox_action
 
   BALANCE_EFFECTS = [
-    DEBIT  = :debit,
-    CREDIT = :credit
+    DEBIT  = 'debit',
```
kyaroch: Changing these to strings because they'll become strings when they get saved to the DB. This way we can use them for equality testing in specs.

PR #111, opened 2019-06-12
tab styling

madisoncarr:
My little contribution to the project! Any suggestions on how to center the text in there?


app/views/dashboard/_lockbox_partner.html.erb
```diff
@@ -1,5 +1,9 @@
 <div class="lockbox-partner clearfix">
   <div>
+  <ul class='partner-tabs'>
```
scottschup: It doesn't really matter to the browser, but for human readability, I'd indent the `<ul>` tag 2 spaces. 

If it looks right to you locally, but is appearing differently on Github, it's probably a tabs vs. spaces issue. If so, you should be able to change the settings in your text editor to  use spaces instead of tabs when you hit the tab key. I think most of us are doing tab = 2 spaces.

scottschup:
Is there an issue this relates to? Or a screen shot of what part of the app this is affecting?

app/views/dashboard/_lockbox_partner.html.erb
```diff
@@ -1,5 +1,9 @@
...
+      <li>Action Needed</li>
+      <li>All Activity</li>
```
BintLopez: Let's go ahead and make this links with href="TODO" since we'd need to implement them. That way we can suss out any link related styling issues too while we're addressing the CSS

app/views/dashboard/_lockbox_partner.html.erb
BintLopez: ^^ yup, agree with this comment

BintLopez:
When I logged in, I noticed that there was a space where an alert would have been displayed that appeared and disappeared between the header nav and these links (alerts aren't currently 100% working but they're working enough to show that an alert would have been in an empty space).

I think it might make more sense to either 1) move these links to the header partial OR 2) move the errors to a different place instead of where they are in [views/layouts/application.html.erb](
https://github.com/MidwestAccessCoalition/lockbox_rails/blob/master/app/views/layouts/application.html.erb#L14-L24). I think I favor the first route since these quick links act more like app navigation stuff.

You could move the code for these into it's own partial and include it in the header partial... https://github.com/MidwestAccessCoalition/lockbox_rails/blob/master/app/views/layouts/_header.html.erb ... I'd recommend naming it something like secondary-nav or content-nav. Since this will make this secondary nav always appear, I'd add an if statement around it inside the header partial that will render it if the user is a partner. Please lmk if that makes sense and @scottschup chime if you think I'm off base here.

app/views/dashboard/_lockbox_partner.html.erb
madisoncarr: Thanks, Scott and Nicole! I'll go ahead and incorporate these changes and will move the code to its own partial so we can render it correctly - thanks so much for taking a look at this!

BintLopez:
image

image

image

Issue #117, opened 2019-06-16
Support Requests should be nested under a Lockbox Partner


BintLopez:
- all routes for support requests should be nested under its lockbox partner

PR #125, opened 2019-06-16
nest support request controller

BintLopez:
https://github.com/MidwestAccessCoalition/lockbox_rails/issues/117


app/controllers/support_requests_controller.rb
```diff
@@ -1,51 +1,18 @@
 require './lib/create_support_request'
 
-class SupportRequestsController < ApplicationController
+class SupportRequestsController < LockboxPartners::SupportRequestsController
```
BintLopez: I'm inheriting to reduce the duplication

BintLopez: I tried to keep it to one controller with custom routing... but had to fight rails too much.

app/controllers/lockbox_partners/support_requests_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+    @support_request = SupportRequest.includes(:notes).find(params[:id])
+  end
+
+  private
+
+  def all_support_request_params
+    support_request_params
+      .merge(lockbox_action: lockbox_action_params)
+      .merge(user_id: current_user.id)
+      .merge(lockbox_partner_id: params[:lockbox_partner_id])
```
kyaroch: It looks to me like if both the form and the URL param provide different `lockbox_partner_id`s, the URL param will take precedence. This is the opposite of what I'd expect to happen. I realize that field is not rendered on the form in that case, so it currently can't happen, but it might make sense to point this out in a comment somewhere in case we create a new form later.

BintLopez: good looking out -- thanks!

Issue #107, opened 2019-06-09
Add notes to support requests


BintLopez:
As a user (admin/coordinator _and_ lockbox partner users), from the page of a support request, I should be able to add a note to a support request.

BintLopez:
https://github.com/MidwestAccessCoalition/lockbox_rails/pull/104#discussion_r294098363

PR #126, opened 2019-06-23
Support request notes 2

misternu:
Alerts do not fade, to test this you need to manually go to a url such as `/support_requests/1`

### DIFF
--- a/app/controllers/support_requests_controller.rb
+++ b/app/controllers/support_requests_controller.rb
@@ -1,8 +1,6 @@
 require './lib/create_support_request'
 
 class SupportRequestsController < ApplicationController
-  def index; end
-
   def show
     @support_request = SupportRequest.includes(:notes).find(params[:id])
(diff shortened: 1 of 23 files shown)


misternu:
#107

app/controllers/support_requests_controller.rb
```diff
@@ -1,6 +1,12 @@
 require './lib/create_support_request'
 
 class SupportRequestsController < ApplicationController
+  def index; end
```
BintLopez: let's rm this since it's not used

db/schema.rb
```diff
@@ -29,6 +29,7 @@
 
   create_table "lockbox_partners", force: :cascade do |t|
     t.string "name"
+    t.string "address"
```
BintLopez: Rebase again and remigrate -- this field shouldn't be added to the schema

Issue #114, opened 2019-06-16
Make alerts visible


BintLopez:
Alerts don't currently appear when login fails or when action are successful. Switching to bootstrap broke this.

PR #166, opened 2019-08-04
fade alerts and dont show error on loading index when logged out

misternu:
#114 make alerts visible

I made sure to keep Scott's alert customizations for the support request show by bumping their specificity


BintLopez:
Alerts on the log in screen are overlaying the text
image

Issue #116, opened 2019-06-16
Fix placement of remember me login box


BintLopez:
<img width="644" alt="image" src="https://user-images.githubusercontent.com/5728859/59568659-4832ad00-9043-11e9-8d16-7c4fb0df4a7b.png">

PR #167, opened 2019-08-04
Fixes placement of remember me

hollielu:
#116 Fix placement of remember me login box 

Now looks like this. ⬇️

Screen Shot 2019-08-04 at 3 11 54 PM

The 100% width was what was making the checkbox jump around crazily when the window was resized, please double-check to make sure removing that won't break anything! 🙂

### DIFF
--- a/app/assets/stylesheets/devise.scss
+++ b/app/assets/stylesheets/devise.scss
@@ -25,9 +25,6 @@ html, body {
     width: 400px;
     margin: 6em auto;
     height: 100%;
-    input {
-      width: 100%;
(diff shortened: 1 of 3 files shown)


BintLopez:
Yes, per @misternu, this should be good to go once the following edits are made:

- move the forgot password? link up so that it's right under the password input
- fix the mobile display such that there are margins

Note -- the screenshot from Jane.0 app's log in screen does show a link for sign up. The Lockbox app is invite-only and so the sign up functionality does not apply here.

Issue #135, opened 2019-07-07
Build out lb partner action needed display


BintLopez:
As a lockbox partner user, when I log in, I should see the action needed tab with all the pending cases showing ordered by pickup date (oldest to latest by pickup date)

BintLopez:
https://github.com/MidwestAccessCoalition/lockbox_rails/pull/131

jel535s:
Time estimate: should be resolved today 8/25

PR #181, opened 2019-08-18
Build out lb partner views

BintLopez:
##  Changelog
- Built out the action items (pending support requests) view for partners
- Remove breakdown and unnecessary details from the partner support request view

## Link to issue:  
- https://github.com/MidwestAccessCoalition/lockbox_rails/issues/135
- https://github.com/MidwestAccessCoalition/lockbox_rails/issues/182

- N/A

## Relevant Screenshots: 
image

image

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/app/models/support_request.rb
+++ b/app/models/support_request.rb
@@ -34,6 +34,10 @@ class SupportRequest < ApplicationRecord
     lockbox_action.eff_date
   end
 
+  def most_recent_note
+    @most_recent_note ||= notes.last
+  end
+
   private
 
   def populate_client_ref_id
--- a/app/models/support_request.rb
+++ b/app/models/support_request.rb
@@ -1,7 +1,7 @@
 class SupportRequest < ApplicationRecord
   belongs_to :lockbox_partner
   belongs_to :user
-  has_many :lockbox_actions
+  has_one :lockbox_action
   has_many :notes, as: :notable
 
   validates :client_ref_id, presence: true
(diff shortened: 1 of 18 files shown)


BintLopez:
Current progress:
image

Todos

-[ ] need to fix notes so that author/commenter shows up
-[ ] need to add hyperlink to go to the support request view
-[ ] make the pending button a dropdown

app/models/support_request.rb
```diff
@@ -12,6 +12,12 @@ class SupportRequest < ApplicationRecord
   # Sometimes the UUID will already have been created elsewhere, and sometimes not
   before_validation :populate_client_ref_id
 
+  def self.pending_for_partner(lockbox_partner_id:)
+    LockboxAction.where.not(support_request_id: nil)
+      .where(lockbox_partner_id: lockbox_partner_id, status: LockboxAction::PENDING)
+      .map(&:support_request)
```
kyaroch: Won't this potentially return duplicates? I don't think there's currently a situation where a support request can have multiple lockbox actions, but there's no actual constraint against that.

(Also, it's possible to do this all in one query using an inner join, but I would guess the performance won't be too bad anyway since there won't be a large number of pending support requests.)

BintLopez: going to change up the model per our discussion so it's only a 1-1 relationship

BintLopez:
image

Issue #179, opened 2019-08-18
As a partner, I should not have god view mode.


rebeldroid12:
## Context
Currently when you log in as a partner you have the options a mac user has. Partners should not have the options to `view all lockboxes` or `file a support request` or `add a new lockbox partner`. It shouldn't have any quick actions.

## Process
Log into the app as a partner user and hit the hamburger

## Acceptance Criteria/Definition of Done
No quick actions available to a partner user. (but keep that for the mac user)

## Resources
![godmodeview](https://user-images.githubusercontent.com/5873894/63230947-250ab200-c1da-11e9-96b1-4c10e7becb75.png)

PR #186, opened 2019-08-25
Restrict partner access

BintLopez:
## Changelog
- Restrict partner access to select views

## Link to issue:  
https://github.com/MidwestAccessCoalition/lockbox_rails/issues/179

- Need to add some tests

## Relevant Screenshots: 
- n/a

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,17 +1,17 @@
 class ApplicationController < ActionController::Base
   before_action :authenticate_user!
 
-  def ensure_admin_only!
-    if !current_user.admin? && current_user.lockbox_partner != @lockbox_partner
-      flash[:error] = "You are not authorized to access this page"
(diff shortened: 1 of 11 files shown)


app/controllers/application_controller.rb
```diff
@@ -1,3 +1,9 @@
 class ApplicationController < ActionController::Base
   before_action :authenticate_user!
+
+  def ensure_admin_only!
+    if !current_user.admin? && current_user.lockbox_partner != @lockbox_partner
```
misternu: A user can bypass the second condition if @lockbox_partner is nil and their lockbox partner is not set, not sure if that can necessarily happen.

This might be easier to read:
```
return if current_user.admin?
return if #other condition tbd
redirect_to root_path
```

Issue #28, opened 2019-04-21
Notify MAC finance users when lockbox amount drops below $300


kyaroch:
When the balance of any lockbox drops below $300, MAC finance users should be notified so that more money can be added to the lockbox.

BintLopez:
Send email notification to mac users when a given lockbox partner balance drops below $300

nsiwnf:
Check balance after lockbox transaction - if below $300, send email

BintLopez:
This should be implemented as an observer/listening/callback action upon balance changing to below $300

Any subsequent support request that changes the balance further below $300 should send additional alerts to the finance team

arches:
What email should be notified?

jel535s:
Time estimate: 8 hours
Potential messaging:
"Heads up!  The lockbox at [partner name] is below $300 when taking into account all pending support requests.  Please send a check ASAP."  
I am also adding a question to my Veronica-user-test script, to specifically note if this messaging needs to be more specific.

PR #188, opened 2019-08-26
send an email when lockbox balance drops below $300

arches:
## Changelog
- Mailer for low balance alerts
- `low_balance?` method on `LockboxPartner`
- Hook to send the email when a `SupportRequest` brings the balance below the threshold

## Link to issue:  
Fixes issue #28 

- N/A

## Relevant Screenshots: 
Example email:
_LOW_LOCKBOX_BALANCE__Owl_Health_Associates_needs_cash

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue

## Deploy

Requires setting the `LOW_BALANCE_ALERT_EMAIL` env var

### DIFF
--- a/app/mailers/lockbox_partner_mailer.rb
+++ b/app/mailers/lockbox_partner_mailer.rb
@@ -1,0 +1,8 @@
+class LockboxPartnerMailer < ApplicationMailer
+  def low_balance_alert
+    return unless ENV['LOW_BALANCE_ALERT_EMAIL'].present?
+
+    @lockbox_partner = params[:lockbox_partner]
+    mail(to: ENV['LOW_BALANCE_ALERT_EMAIL'], subject: "[LOW LOCKBOX BALANCE] #{@lockbox_partner.name} needs cash")
+  end
+end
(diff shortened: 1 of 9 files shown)


app/mailers/lockbox_partner_mailer.rb
```diff
@@ -0,0 +1,8 @@
+class LockboxPartnerMailer < ApplicationMailer
+  def low_balance_alert
+    return unless ENV['LOW_BALANCE_ALERT_EMAIL'].present?
+
+    @lockbox_partner = params[:lockbox_partner]
+    mail(to: ENV['LOW_BALANCE_ALERT_EMAIL'], subject: "[LOW LOCKBOX BALANCE] #{@lockbox_partner.name} needs cash")
```
BintLopez: `LOW_BALANCE_ALERT_EMAIL` should be able to handle multiple emails (comma separated)

arches: Oh yeah I had a hunch about that one but hadn't actually looked into it, thanks for reminding me... upon some research I'm pretty sure ActionMailer will accept a comma-separated `to` value. https://guides.rubyonrails.org/action_mailer_basics.html#sending-email-to-multiple-recipients

When I use a `LOW_BALANCE_ALERT_EMAIL` with a comma it still generates a single email that includes both addresses in the `to` field, which _looks_ right... but I haven't figured out a good way to truly verify it except doing it live and seeing if it shows up in both inboxes, which I don't have a way to do at the moment.  Do we have a staging environment with live emails?

Issue #155, opened 2019-07-28
Add all support requests view for MAC users


BintLopez:
<img width="1121" alt="image" src="https://user-images.githubusercontent.com/5728859/62013270-987a5000-b155-11e9-98a4-6c93fd8afe18.png">

## Questions for user testing
* which support requests do we want to show in this view (only pending?)
* do we only want to the most recent X support requests?
* is it useful to see the balance here?

jel535s:
This seems to be a duplicate of https://github.com/MidwestAccessCoalition/lockbox_rails/issues/61

I've also added these questions into the documentation for issue #156

BintLopez:
Note -- the word "Clinic" is used in this table in the mockup. However, let's go with the term "Partner" as not all lockbox partners will necessarily be a clinic.

jel535s:
Time estimate: 8 hours

PR #189, opened 2019-08-28
support requests index view

arches:
## Changelog
- status scopes on support request
- support requests index page
- dropdown to switch between pages\

## Link to issue:  
Fixes issue #155 

- view should only be available for admins

## Relevant Screenshots: 
New view at `/support_requests`

LockboxRails

## Caveats

- I'm not sure this select-based nav is going to be very extensible, pretty soon we're going to want to filter within the support requests index instead of switching to a new URL.
- I'm not sure if there are any other mocks with implications on this select either.
- No limit on the number of support requests we show.  Probably fine for quite a while, but if this starts timing out we'll need to paginate.
- I didn't include a count in the dropdown... I've found counts in nav items to be frustrating over time.
- Also I used a select instead of a link, I'm not sure how closely we're trying to hit these mocks
- Flash messages aren't showing up

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/app/assets/stylesheets/support_request.scss
+++ b/app/assets/stylesheets/support_request.scss
@@ -1,3 +1,12 @@
+#support-requests-index {
+  max-width: 600px;
+  margin: 0 auto;
+
+  h2 {
+    margin: 30px auto;
+  }
+}
+
 .support-request-header {
   margin: 30px auto;
 }
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,3 +1,10 @@
 class ApplicationController < ActionController::Base
   before_action :authenticate_user!
+
+  def require_admin
+    unless current_user.admin?
+      flash[:error] = "You are not authorized to access this page"
+      return redirect_to root_path
+    end
+  end
 end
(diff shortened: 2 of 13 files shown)


BintLopez:
Thoughts re some of the caveats...
> Flash messages aren't showing up

^^ @misternu -- thinking your PR will fix this, lmk if I'm wrong here

> - I didn't include a count in the dropdown... I've found counts in nav items to be frustrating over time.
> - Also I used a select instead of a link, I'm not sure how closely we're trying to hit these mocks

^^ Let's log a follow up issue for the styling so we remember to address later and note the count cc/ @gujacq for visibility

> - I'm not sure this select-based nav is going to be very extensible, pretty soon we're going to want to filter within the support requests index instead of switching to a new URL.
> - No limit on the number of support requests we show. Probably fine for quite a while, but if this starts timing out we'll need to paginate.

^^ I think this is a good first step and we'll cross the pagination bridge and/or filtering from the support request index when we get there

arches:
> Thoughts re some of the caveats...
> 
> > Flash messages aren't showing up
> 
> ^^ @misternu -- thinking your PR will fix this, lmk if I'm wrong here

yes! I was excited to see that PR, it will show the flash.  I didn't see it styling `flash[:error]` though so it could use a quick visual touch-up now that we have this use case.

Issue #29, opened 2019-04-21
Clinic user can update the status of a support request.


kyaroch:
As a lockbox partner user, I can update the status of a support request from pending to completed or canceled.

<img width="1195" alt="image" src="https://user-images.githubusercontent.com/5728859/58386463-bf859b80-7fc5-11e9-8f53-0aebd661182f.png">

PR #192, opened 2019-09-01
Add support request status update dropdown

misternu:
## Changelog
* Added a status update dropdown to the lockbox_partner/support_request#show page
* Duplicated dropdown to the pending support request partial in action needed view

## Link to issue:  
#29 

## Relevant Screenshots: 
Screen Shot 2019-09-01 at 3 11 09 PM
Screen Shot 2019-09-01 at 3 11 21 PM

Note, because these dropdowns are shared between two views, both redirect to the support request show page when completed. If this is not the desired ux it can be easily changed

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -1,6 +1,5 @@
 class DashboardController < ApplicationController
   def index
-    flash[:notice] = "This is a test"
     if current_user.partner?
       @lockbox_partner = current_user.lockbox_partner
     else
--- a/app/controllers/lockbox_partners/support_requests_controller.rb
+++ b/app/controllers/lockbox_partners/support_requests_controller.rb
@@ -29,6 +29,18 @@ class LockboxPartners::SupportRequestsController < ApplicationController
     @lockbox_partner = @support_request.lockbox_partner
   end
 
+  def update_status
+    @support_request = SupportRequest.find(params[:support_request_id])
+    @lockbox_partner = @support_request.lockbox_partner
+    status = update_status_params[:status]
+    if @support_request.lockbox_action.update(status: status)
+      flash[:notice] = "Status updated to #{status}"
+    else
+      flash[:error] = "Failed to update status"
+    end
+    redirect_to lockbox_partner_support_request_path(id: @support_request.id)
+  end
+
   private
 
   def all_support_request_params
@@ -45,6 +57,10 @@ class LockboxPartners::SupportRequestsController < ApplicationController
       :urgency_flag,
       :lockbox_partner_id
     )
+  end
+
+  def update_status_params
+    params.permit(:status)
   end
 
   def lockbox_action_params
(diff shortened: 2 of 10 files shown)


app/models/support_request.rb
```diff
@@ -37,6 +37,16 @@ def most_recent_note
...
   end
 
+  def status_options
+    LockboxAction::STATUSES - [status]
+  end
+
+  def update_status_path(params)
+    params = { lockbox_partner_id: lockbox_partner_id, support_request_id: id }.merge(params)
+    Rails.application.routes.url_helpers
+      .lockbox_partner_support_request_update_status_path(params)
```
arches: Routes in a model are generally considered a code smell, because the models are intended to represent business logic but routes are particular to the workflows we're building on top of that business logic. I think there's a way to construct the view using the routes directly. And if not, I usually see this type of thing inside a literal helper file (or a presenter).

app/controllers/dashboard_controller.rb
```diff
@@ -1,6 +1,5 @@
 class DashboardController < ApplicationController
   def index
-    flash[:notice] = "This is a test"
```
arches: 👍👍

app/controllers/lockbox_partners/support_requests_controller.rb
```diff
@@ -29,6 +29,18 @@ def show
     @lockbox_partner = @support_request.lockbox_partner
   end
 
+  def update_status
```
arches: How might this look as a special case of an `update` flow rather than its own flow? Any tradeoffs one way or the other?

app/models/support_request.rb
```diff
@@ -37,6 +37,16 @@ def most_recent_note
     @most_recent_note ||= notes.last
   end
 
+  def status_options
+    LockboxAction::STATUSES - [status]
```
arches: nice 🙂

app/controllers/lockbox_partners/support_requests_controller.rb
```diff
@@ -29,6 +29,18 @@ def show
...
+    @support_request = SupportRequest.find(params[:support_request_id])
+    @lockbox_partner = @support_request.lockbox_partner
+    status = update_status_params[:status]
+    if @support_request.lockbox_action.update(status: status)
+      flash[:notice] = "Status updated to #{status}"
+    else
+      flash[:error] = "Failed to update status"
```
arches: 👌

app/views/lockbox_partners/support_requests/_status_dropdown.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  </a>
+  <div class="dropdown-menu" aria-labelledby="dropdownMenuLink">
+    <% support_request.status_options.each do |option| %>
+      <%= link_to option.capitalize,
+        lockbox_partner_support_request_update_status_path(
+          support_request_id: support_request.id,
+          lockbox_partner_id: support_request.lockbox_partner_id,
+          status: option
+        ),
+        class: 'dropdown-item', method: 'post', remote: true %>
```
arches: wow that's all it takes?!

misternu: 39j7rg

Issue #184, opened 2019-08-22
As a partner, I need a confirmation of deposit to be sent so I can confirm.


rebeldroid12:
## Context
Coordinators can add cash to the partners but there is message asking to confirm the deposit. 

## Process
As a coordinator, add cash to a partner. As a partner, look at the logs. The total is updated and there is nothing for a partner to confirm the deposit. So technically, this could lead to the situation where no money was received but our system says it was. 

## Acceptance Criteria/Definition of Done
Trigger a confirmation of deposit to be sent when cash has been added by the coordinator. This message will alert the partner to be on the look out for the deposit in the mail. 

## Resources
https://projects.invisionapp.com/share/9EMYOUG7ZBA#/screens/359242753

BintLopez:
When a MAC user adds cash to a lockbox...
- the lockbox partner should receive a notification email that cash is on the way
- an alert should appear on the lockbox partner's screen prompting them to confirm the cash has been received

PR #203, opened 2019-09-09
partners can confirm deposit

arches:
Ability for partners to see that cash is on the way, and confirm they received it.

## Changelog
- New controller for working with LockboxActions
- New predicate method `LockboxPartner#awaiting_infusion?`
- New alert for partners

## Link to issue:  
Fixes issue #184 

- N/A

## Open Questions

- [ ] Is `eff_date` the date the money was sent?
- [ ] Are all infusions checks?

## Relevant Screenshots: 

Normally you'd only have one of these, but they'll stack up until they're confirmed.

alerts

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/app/controllers/lockbox_actions_controller.rb
+++ b/app/controllers/lockbox_actions_controller.rb
@@ -1,0 +1,23 @@
+class LockboxActionsController < ApplicationController
+  before_action :find_lockbox_action, only: [:update]
+
+  def update
+    if @lockbox_action.update(update_params)
+      flash[:notice] = "Success!"
+      redirect_to lockbox_partners_path
+    else
+      flash[:error] = "Sorry, there was a problem."
+      redirect_to lockbox_partners_path
+    end
+  end
+
+  private
+
+  def find_lockbox_action
+    @lockbox_action = LockboxAction.find(params[:id])
+  end
+
+  def update_params
+    params.require(:lockbox_action).permit(:status)
+  end
+end
--- a/app/controllers/lockbox_actions_controller.rb
+++ b/app/controllers/lockbox_actions_controller.rb
@@ -1,5 +1,6 @@
 class LockboxActionsController < ApplicationController
   before_action :find_lockbox_action, only: [:update]
+  before_action :require_ownership
 
   def update
     if @lockbox_action.update(update_params)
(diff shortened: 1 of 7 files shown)


app/controllers/lockbox_actions_controller.rb
```diff
@@ -0,0 +1,30 @@
+class LockboxActionsController < ApplicationController
+  before_action :find_lockbox_action, only: [:update]
+  before_action :require_ownership
+
+  def update
+    if @lockbox_action.update(update_params)
+      flash[:notice] = "Success!"
+      redirect_to lockbox_partners_path
+    else
+      flash[:error] = "Sorry, there was a problem."
```
BintLopez: need to use 'alert' instead of error to get the alert styling (or add styling for error vs alert messages)

app/models/lockbox_partner.rb
```diff
@@ -35,6 +35,10 @@ def low_balance?
     balance < MINIMUM_ACCEPTABLE_BALANCE
   end
 
+  def awaiting_infusion?
```
BintLopez: alternate naming suggestion `awaiting_cash_confirmation?`

arches: hm actually now that I think about it, `cash_confirmation` isn't very description.  `awaiting_cash_addition_confirmation` would be more closely aligned with the `AddCashToLockbox` interactor naming, but kind of awkward.

arches: The partner is waiting on the cash, _we_ are waiting on the confirmation.  In general I prefer to name from the model's POV, so including "confirmation" in the name doesn't quite resonate with me.

`awaiting_cash_infusion?`
`awaiting_cash_delivery?`
`awaiting_cash_addition?`
`awaiting_cash_distribution?`
`awaiting_cash?`
`transfer_in_progress?` <-- doesn't imply a credit, not sure if that's good/bad/neutral

BintLopez: What about `cash_addition_received?` and then we'd check for the negative?

BintLopez: actually ignore that one

BintLopez: `awaiting_cash_received_confirmation?`

arches: Notes from offline: we've had trouble in the past getting them to submit a form when they receive the cash, so the workflow we're creating here is more about us checking up on the partner, vs the partner managing their own balances.  That means a partner-based POV doesn't resonate, nor do partner-based names.  So instead we intend to go with a name that's more closely aligned with our POV on the state of the partner, e.g. `cash_addition_confirmation_pending?`

Issue #113, opened 2019-06-16
Fix behavior on successful add cash submit


kyaroch:
- When a MAC user successfully adds cash to the lockbox, the form currently redirects to the lockbox partner view and doesn't show any message to the user. We need to decide what the behavior should be and then implement that behavior.
- Currently, when a user navigates to the add cash form from a lockbox partner, the lockbox partner field defaults to the selected lockbox partner but can be edited. We need to decide whether this field should be disabled or not.
- If the user selects a different lockbox partner, they will still be redirected to the lockbox partner they were originally viewing upon successful submission. This is probably not what we want.

gujacq:
Here is the confirmation screen we talked about: https://invis.io/9EMYOUG7ZBA#/370234473_dashboard_-_MAC_User_-cash_Added_Confirmation-

We also decided that we would let users reselect that lockbox partner... but that if they switched lockbox partners we'd make sure the information is tied to the correct partner in the backend.

gujacq:
Kevin might be working on this already...

jel535s:
This design looks good, and is ready to work.  

Story:
As a MAC user, when I successfully add cash to the lockbox, the form currently redirects to the lockbox partner view and doesn't show any message to the user.  I want to be shown a confirmation screen on that redirect. (see design example above).

Also, as a MAC user, when I navigate to the add cash form from a lockbox partner's page, I want to be able to reselect that lockbox partner on the add cash page and have the information tied to the correct partner in the backend.  Currently, when a user navigates to the add cash form from a lockbox partner, the lockbox partner field defaults to the selected lockbox partner but can be edited.

If the MAC user selects a different lockbox partner on the add cash form, they should be redirected to the new lockbox partner that they selected (when cash is successfully added, the user screen redirects to the lockbox partner field where the cash was just added).

arches:
> We also decided that we would let users reselect that lockbox partner... but that if they switched lockbox partners we'd make sure the information is tied to the correct partner in the backend.

😬that's totally fine with me but we can't use a URL like `/lockbox_partners/1` to add cash to partner 2.  That's how it's currently working, so I'll reset the URLs as part of this ticket.

PR #206, opened 2019-09-15
rearrange cash addition workflow

arches:
As a MAC user, when I successfully add cash to the lockbox, the form currently redirects to the lockbox partner view and doesn't show any message to the user. I want to be shown a confirmation screen on that redirect.

## Changelog
* Since the user can change the partner in the form, move the `add_cash` route out of the partner resource so the form path doesn't require and ID
* Redirect to the selected partner upon successful submission
* Add success/failure messages

## Link to issue:  
Fixes issue #113 

- N/A

## Relevant Screenshots: 

## Are you ready for review?:

- [x] ~Added relevant tests~
- [x] Linked PR to the issue
- [x] ~Added notes for QA/special notes~
- [x] ~Added revelant screenshots~

### DIFF
--- a/app/controllers/lockbox_partners/add_cash_controller.rb
+++ b/app/controllers/lockbox_partners/add_cash_controller.rb
@@ -2,7 +2,7 @@ require 'add_cash_to_lockbox'
 
 class LockboxPartners::AddCashController < ApplicationController
   before_action :set_lockbox_partner, only: :create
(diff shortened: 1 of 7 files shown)


BintLopez:
As discussed, let's axe the ability to change which partner we're adding cash to from the add cash page. This option makes the routing a bit confusing and inconsistent, so better to not allow this option at all.

PR #240, opened 2019-10-15
I think this is a typo should be "td" instead of "th"

nsiwnf:
Fixing typo


nsiwnf:
Please feel free to merge if this is in fact a typo

Issue #140, opened 2019-07-07
Thoroughly QA reconciliation


BintLopez:
Currently seeing some interesting behavior around reconciliation of lockboxes, especially when there are pending tasks present

jel535s:
Notes from 8/25 meeting: needs test scenarios.  Bug has shown up in Reconciliation.  Balance results don’t match when we create fake support requests.  Balance on Reconciliation might factor in Pending requests when it shouldn’t be.

- Designing test cases
- Getting someone set up to do manual testing
- Doing manual testing
- Creating and running automated tests

Time estimate: 24 hours

BintLopez:
There is a bug with reconciliation where a lockbox transaction with a negative amount can be created. This should not be the case.

PR #244, opened 2019-10-27
fix negative recon amounts

AG-Sile:

## Changelog
- added validation for lockbox_transaction to never be below zero
- Set Reconciliation to always post absolute value of the difference when creating a lockbox_transaction

## Link to issue:  
Fixes issue https://github.com/MidwestAccessCoalition/lockbox_rails/issues/140

- This bug would occur when the recon amount was less than the current amount in the lockbox, it would then create negaitve lockbox_transactions which resulted in most calculations being incorrect.

## Relevant Screenshots: 

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/spec/lib/reconciliation_spec.rb
+++ b/spec/lib/reconciliation_spec.rb
@@ -30,6 +30,32 @@ describe Reconciliation do
         .and_return(Monetize.parse('1000.00'))
     end
 
+    context 'when the amount is less than lockbox balance' do
+      let(:amount) { Money.new(900_00) }
+
+      it 'creates one lockbox action' do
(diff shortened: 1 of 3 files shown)


spec/lib/reconciliation_spec.rb
```diff
@@ -30,6 +30,32 @@ def reconcile
         .and_return(Monetize.parse('1000.00'))
     end
 
+    context 'when the amount is less than lockbox balance' do
```
BintLopez: 'when the actual counted amount is less than the expected lockbox balance'

BintLopez: pls rephrase and then this is good to merge!

Issue #204, opened 2019-09-15
Implement MAC user global forms header


BintLopez:
As a MAC user, when I'm not looking at a particular lockbox and I click on either the "File Support Request" or "Add Lockbox" buttons I should see the forms to do either of those things under the buttons. The button that I clicked should have an indicated notch (little triangle underneath it) to indicate which form I'm on.

See:  
https://projects.invisionapp.com/share/9EMYOUG7ZBA#/screens/365424955
https://projects.invisionapp.com/share/9EMYOUG7ZBA#/screens/365424956

## Acceptance Criteria
- Support Request and Add a new lockbox forms from the MAC user dashboard should match the mockups in the links above.

nsiwnf:
This is what I see right now - so should these tabs initially be "closed" and then when the user clicks on it, it will show whats in the mocks including all the support requests that are there right now? 
<img width="695" alt="Screen Shot 2019-10-19 at 11 00 12 AM" src="https://user-images.githubusercontent.com/34173394/67149391-a70d6a80-f25f-11e9-86a6-3e98699d47bc.png">

BintLopez:
@nsiwnf ^^ yes! That is correct! They're in 'closed' state when we're looking at all the support requests or all the lockbox partners, and have the triangle only when either filing a support request or creating a new lockbox is happening.

gujacq:
The wifi situation is failing me so I can't update Invision at the time but here's as screenshot of the back to dashboard link

<img width="1053" alt="Screen Shot 2019-11-10 at 5 32 54 PM" src="https://user-images.githubusercontent.com/33872753/68552760-544b4c80-03e0-11ea-80bc-222f5761026c.png">

nsiwnf:
@gujacq @BintLopez I wasn't able to get the triangle to have a border, but I did get just a blue triangle. Should I move forward with this or just leave it without the triangle?
<img width="626" alt="Screen Shot 2019-11-30 at 1 57 41 PM" src="https://user-images.githubusercontent.com/34173394/69906519-8dd8fd00-1379-11ea-8a81-27d4ef760a72.png">

BintLopez:
I personally am cool with this to get us to launch and then we can go back and revise/tweak the styles later. @gujacq you cool with that?

PR #255, opened 2019-11-06
#204 Implement MAC user global forms header

nsiwnf:

## Changelog
- Added a global header for forms

## Link to issue:  
#204 

- Cosmetic change 

## Relevant Screenshots: 
On login :
Screen Shot 2019-11-06 at 1 36 16 AM

Click on "Create a new lockbox partner" 
Screen Shot 2019-11-29 at 8 07 13 PM

Click on "File a support request"
Screen Shot 2019-11-29 at 8 06 56 PM

## Are you ready for review?:

- [ ] Added relevant tests
- [x] Linked PR to the issue


app/assets/stylesheets/bootstrap-overrides.scss
```diff
@@ -30,6 +30,21 @@ a {
...
+  top: auto;
+  left: 0;
+  right: 0;
+  margin: 0px 0px 0px 60px;
+  width: 0;
+  height: 0;
+  border-top: solid 10px $color-primary;
+  border-left: solid 10px transparent;
+  border-right: solid 10px transparent;
+}
```
nsiwnf: I can't get the little triangles to display properly

BintLopez: Will sync up with @gujacq to figure out a workaround styling if the triangles are a no-go.

BintLopez:
Looked over this with @gujacq ... a little bit of feedback re UI stuff....

- The form should not have margins/padding on the side but should rather be at 100% to match the width of the top buttons and line
- For "Create new lockbox partner", the form title should match the button text
image

BintLopez:
We can scrap the triangles for now -- going to add a follow up task to add these. This is good to merge once the forms are at full width and the title of the new partner form matches the text of the button.

BintLopez:
Jk.. a few more things here...

1. per @gujacq, we should keep the "Welcome {NAME}" in the header for these forms to match the header when looking at all the lockboxes.

2. We need to add a link "Back to Dashboard" from these forms so that coordinators can easily get back to the All lockboxes views (@gujacq is updating the issue to show a mockup)

3. It looks like the stylings applied here are also affecting the headers for a specific lockbox
image

nsiwnf:
Quick updates/fixes :
Screen Shot 2019-11-24 at 12 27 34 PM

(no longer being squished)
Screen Shot 2019-11-24 at 12 29 30 PM

nsiwnf:
Also fixed styling
Screen Shot 2019-11-29 at 8 07 32 PM

app/views/lockbox_partners/_header.html.erb
```diff
@@ -1,7 +1,7 @@
 <div class="lockbox-partner-header">
   <strong><%= lockbox_partner.city %></strong>
   <% if current_user.admin? %>
-    <%= link_to 'Back to dashboard', lockbox_partners_path, class: 'float-right' %>
```
BintLopez: We need to preserve the `float-right` for the lockbox partner specific dashboard view. If we float left, the below display gets weird instead of being over the amount. Could potentially pass in the desired alignment to the partial and then interpolate `class: "float-#{alignment}"`

image

nsiwnf: Ah got it --->
Screen Shot 2019-11-30 at 11 51 01 AM

BintLopez:
The back to dashboard links look look good as do the forms. Two pieces of very minor feedback related to the "Add a new lockbox partner" view and then this is good to merge imo...

1. unlike the "File a support request" form, this form still has 15px padding and so is not spanning the full width like the file support request form does. I think this is being inherited from the `container` class
2. Let's change the phrasing on the button to match the form title "Create new lockbox partner" --> "Add a new lockbox partner"

image

BintLopez:
Re your [comment on the issue](https://github.com/MidwestAccessCoalition/lockbox_rails/issues/204#issuecomment-560024654)... I'm good merging it with the blue triangle or just leaving this as a follow up task to solve.

PR #263, opened 2019-11-17
make low balance alert specs more reliable

arches:
## Changelog
- Explicitly set values in specs instead of allowing them to be picked up from local env

## Are you ready for review?:

- [x] Added relevant tests
- ~[ ] Linked PR to the issue~
- ~[~ ] Added notes for QA/special notes
- ~[ ] Added revelant screenshots~

### DIFF
--- a/spec/lib/create_support_request_spec.rb
+++ b/spec/lib/create_support_request_spec.rb
@@ -86,7 +86,7 @@ describe CreateSupportRequest do
     end
 
     it 'goes to the finance team when balance is below $300' do
-      ENV['LOW_BALANCE_ALERT_EMAIL'] = 'lowbalance@alert.com'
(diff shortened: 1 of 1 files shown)


spec/lib/create_support_request_spec.rb
```diff
@@ -86,7 +86,7 @@
     end
 
     it 'goes to the finance team when balance is below $300' do
-      ENV['LOW_BALANCE_ALERT_EMAIL'] ||= 'lowbalance@alert.com'
+      ENV['LOW_BALANCE_ALERT_EMAIL'] = 'lowbalance@alert.com'
```
AG-Sile: consider stubbing out the env vars instead..     allow(ENV).to receive(:[]).with("LOW_BALANCE_ALERT_EMAIL").and_return("lowbalance@alert.com")

arches: :+1: good idea, thanks!

PR #269, opened 2019-11-17
Ensure that CreateSupportRequest doesn't leave DB in inconsistent state 262

### LINKED ISSUE #262: Bug in MAC user all support requests view
Currently, when a MAC user is on the dashboard and selects the view for "all support requests", the link in the row for the support request is broken

image

image

kyaroch:
## Changelog
- Improved error handling in `CreateSupportRequest` action so that transaction is rolled back if any part of it fails
- Added validations to `LockboxTransaction`
- Ensured that transactions are always saved with a category
- Updated specs

## Link to issue:  
Fixes #262 

- Tested support request creation (both through console and actual form) and verified that it still works

## Relevant Screenshots: 
- N/A

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/spec/lib/create_support_request_spec.rb
+++ b/spec/lib/create_support_request_spec.rb
@@ -4,8 +4,18 @@ require './lib/add_cash_to_lockbox'
 
 describe CreateSupportRequest do
   let(:mac_user) { FactoryBot.create(:user) }
+
   let(:lockbox_partner) do
     FactoryBot.create(:lockbox_partner, :active)
+  end
+
+  let(:lockbox_transactions) do
+    [
+      {
+        amount:       42.42,
+        category:     "gas"
+      }
+    ]
   end
 
   let(:params) do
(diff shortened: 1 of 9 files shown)


spec/lib/create_support_request_spec.rb
```diff
@@ -16,27 +26,68 @@
...
+  context "when one of the lockbox transactions is invalid" do
+    let(:lockbox_transactions) do
+      [
+        {
+          amount:   42.42,
+          category: "gas"
+        },
+        {
+          amount:   0.00
+        }
```
BintLopez: In this case, can we also add a transaction with a `category` but `""` for an amount?

spec/models/lockbox_action_spec.rb
```diff
@@ -37,7 +37,7 @@
       FactoryBot.create(:lockbox_action, :support_client).tap do |action|
         action.lockbox_transactions.create(
           amount_cents: 20_00,
-          category: 'gas'
+          category: LockboxTransaction::GAS
```
BintLopez: 🚗

spec/models/lockbox_partner_spec.rb
```diff
@@ -14,7 +14,8 @@ def add_cash(partner, date)
     ).tap do |action|
       action.lockbox_transactions.create!(
         amount_cents: 1000_00,
-        balance_effect: 'credit'
+        balance_effect: 'credit',
```
BintLopez: wanna constantize this one while we're here? `LockboxTransaction::CREDIT`

spec/models/lockbox_partner_spec.rb
```diff
@@ -28,7 +29,8 @@ def pending_request_on(partner, date, amount_breakdown)
       amount_breakdown.each do |amt_cents|
         lb_action.lockbox_transactions.create!(
           amount_cents: amt_cents,
-          balance_effect: 'debit'
+          balance_effect: 'debit',
```
BintLopez: and ditto.... `LockboxTransaction::DEBIT`

spec/lib/create_support_request_spec.rb
kyaroch: In the same test, that wouldn't have any effect, as only one needs to be invalid. I did add another test where both are blank, to verify that we're ignoring that transaction.

PR #283, opened 2019-12-01
Update support request dropdown styling 88

### LINKED ISSUE #88: Update the styling for Support Request status dropdown for MAC users
# CURRENT SPEC:
https://github.com/MidwestAccessCoalition/lockbox_rails/issues/88#issuecomment-522821934

# OLD SPEC (context)
As a MAC user, I may sometimes want to update the amount to be disbursed to a client. Currently when coordinators need to do this, they fill out another lockbox request form for the same client. In the app, this should be handled by the update support request functionality.

If the amount or pickup date changes, the lockbox partner should receive a notification email 

image

kyaroch:
## Changelog
- Revised appearance of support request status dropdown for admin users

## Link to issue:  
Fixes #88 

- N/A

## Relevant Screenshots: 
Screenshot_20191201_183205
Screenshot_20191201_183423

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/app/models/support_request.rb
+++ b/app/models/support_request.rb
@@ -65,7 +65,7 @@ class SupportRequest < ApplicationRecord
   end
 
   def status_options
-    LockboxAction::STATUSES - [status]
+    LockboxAction::STATUSES
   end
 
   def record_creation
--- a/app/models/support_request.rb
+++ b/app/models/support_request.rb
@@ -64,8 +64,12 @@ class SupportRequest < ApplicationRecord
     lockbox_action.status
   end
 
-  def status_options
-    LockboxAction::STATUSES
+  def status_options(include_current: false)
+    if include_current
+      LockboxAction::STATUSES
+    else
+      LockboxAction::STATUSES - status
+    end
   end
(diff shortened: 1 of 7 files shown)


app/models/support_request.rb
```diff
@@ -65,7 +65,7 @@ def status
   end
 
   def status_options
-    LockboxAction::STATUSES - [status]
```
BintLopez: does this mess things up for the partner view when changing statuses?

kyaroch: As far as I can tell, this method isn't called anywhere except in the view I was working on, and in that case we want to display all statuses (including the current one) according to the mockup.

BintLopez:
This broke the update status styling in the "Action Needed" tab for lockbox partner users...

Instead of showing up as it does for MAC users (below)....
image

... it should look like the following (which is current behavior in master)
image

Also... noticed that on hover, the update status button for partners becomes illegible. We should disable the color change on hover for this button.
image

Issue #212, opened 2019-09-16
Build notifications for MAC user/partner: changes to a Support Request


jel535s:
### Context
As a MAC user and Partner, I want to know when  a support request is edited in any of the following ways:
Support Request object:
client alias
client ref ID
urgency flag
OR
Lockbox Action:
pick up date
OR
Lockbox transactions:
-amount breakdown- this will need to be written as "check last amount, if different, write a note"

### Process
Perhaps by incorporating the Papertrail audit gem, build two pieces of notification when a Support Request is created:
1.  System-generated note, visible on the Support Request page when opened by either MAC users or Lockbox partners.  
For edits to the client alias, the note should read: "The client alias for the Support Request was changed on [date + time] by [user] from [previous alias] to [changed alias]."
For edits to the client reference ID, the note should read: "The client reference ID for this Support Request was changed on [date + time] by [user] from [previous ID] to [changed ID]."
For edits to the urgency flag, the note should read: "The urgency flag for this Support Request was changed on [date + time] by [user] from [previous flag] to [changed flag]."
For edits to the pickup date, the note should read: "The pickup date for this Support Request was changed on [date + time] by [user] from [previous date] to [changed date]."
For edits to the total amount, please note, this is more complicated and will need to be to be written as something like, "check last amount, if different, write a note".  This system-generated note should read: "The total amount for this Support Request was changed on [date + time] by [user] from [previous amount] to [changed changed amount]."
Right now, when you create a note without a user, it automatically creates a system-generated note.

2. An email is sent to the Lockbox partner and CC'd to the MAC coordinator.  The email copy reads:
For edits to the client alias, the email should read: 
**Subject:** Support Request [ID] client alias was changed.
**Body:**  The client alias for the Support Request was changed on [date + time] by [user] from [previous alias] to [changed alias].
For edits to the client reference ID, the email should read: 
**Subject:** Support Request [ID] client reference ID was changed.
**Body:**  The client reference ID for this Support Request was changed on [date + time] by [user] from [previous ID] to [changed ID].
For edits to the urgency flag, the note should read: 
**Subject:** Support Request [ID] urgency flag was changed.
**Body:** The urgency flag for this Support Request was changed on [date + time] by [user] from [previous flag] to [changed flag].
For edits to the pickup date, the note should read: 
**Subject:** Support Request [ID] date was changed.
**Body:** The pickup date for this Support Request was changed on [date + time] by [user] from [previous date] to [changed date].
[...]

PR #288, opened 2019-12-08
212 Email Notifications for Support Request Changes

arches:
## Changelog
- New interactor for updating support requests
- Create system note when support request changes
- Integrate existing status change flow into new interactor-based flow
- Send alert emails for "support request changes" (in quotes because really we're sending alert emails on note creation, which is triggered by a support request change)

I think I got it but I'd appreciate some QA  🙏 

## Link to issue:  
Fixes issue #212 

- N/A

## Relevant Screenshots: 

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue


app/mailers/note_mailer.rb
```diff
@@ -0,0 +1,28 @@
...
+  def note_creation_alert
+    @note = params[:note]
+
+    if @note.system_generated?
+      subject = "New System note on #{@note.notable_type.titleize} ##{@note.notable_id}"
+    else
+      subject = "New note from #{@note.author} on #{@note.notable_type.titleize} ##{@note.notable_id}"
+    end
+
+    mail(to: recipients(@note), subject: subject)
```
BintLopez: Let's either change this to bcc or send individual emails to recipients. That way our notifications won't turn into a reply-all thread and we can encourage convos to take place in the notes field.

app/views/note_mailer/note_creation_alert.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  A new note was added to <%= @note.notable_type %> #<%= @note.notable_id %>
+</h1>
+
+<p>
+  <%= @note.text %>
+</p>
+
+<% if @note.notable.class == SupportRequest %>
+  <p>
+    View the note here: <%= lockbox_partner_support_request_url(@note.notable.lockbox_partner, @note.notable) %>
```
BintLopez: Tweak for the phrasing here:  `View and reply to the note here`

lib/update_support_request.rb
```diff
@@ -0,0 +1,83 @@
...
+  #         ...
+  #       },
+  #     }
+  #   },
+  #   name_or_alias: "Foster",
+  #   client_ref_id: "qwertyuiopp",
+  #   urgency_flag: "one"
+  # }
+  #
+  # We'll pass it directly to SupportRequest to take advantage of AcceptsNestedAttributes.
```
BintLopez: 💯

spec/lib/create_support_request_spec.rb
```diff
@@ -188,7 +188,7 @@
       result = nil
 
       expect { result = CreateSupportRequest.call(params: low_balance_params) }
-        .to change{ActionMailer::Base.deliveries.length}.by(2)
+        .to change{ActionMailer::Base.deliveries.length}
```
BintLopez: why was this change necessary?

Gemfile
```diff
@@ -32,6 +32,7 @@ gem 'devise', '~> 4.6.2'
 gem 'font-awesome-rails'
 gem 'font_assets', '>= 0.1.14'
 gem 'bootstrap', '~> 4.3.1'
+gem 'table_print'
```
BintLopez: did you end up using this? If not, let's remove or only have as a dev dependency

lib/update_support_request.rb
arches: Before I knew I could pass an object to `fail` I had to pass in an ActiveRecord object from the controller so after this interactor marked up the ActiveRecord objects with errors I could still access them in the view.  I'll be able to simplify this interactor to `input :params`.

spec/lib/create_support_request_spec.rb
arches: Now that notes send email the count was wrong.  I could have changed it to 3 but since we aren't checking the actual thing that was delivered it didn't seem like the count was adding much.

arches: Or to put it another way.... a change unrelated to this test made the test fail, so I wanted to fix it such that future similar changes would not make the test fail.  E.g. if we split the note alerts out into individual emails per user this could go from 3 to 6 or something.

arches: Though... on further thought.... maybe this entire expectation is not useful (as is), since unrelated changes could make it fail _or pass_ regardless of whether the specific email we care about here is being sent.

Gemfile
arches: I use it all the time, and I find it very helpful in production for debugging, but I'm happy to move it to a dev dependency.

arches: lmk

arches: also check out http://tableprintgem.com :)

BintLopez: hmmmm... I'm wary of having a non 100% necessary dependency in prod for security reasons.... but this does look useful. We can leave it for now and see if we use it for prod things?

spec/lib/create_support_request_spec.rb
BintLopez: So... I took a closer look at this and I don't understand why this test would trigger another note/email for the support request creation. Are we inadvertently triggering an update email for creation? Gonna dive a little deeper and see what's going on...

BintLopez: (this is not a blocker to merging)

arches: This test will trigger three emails:

1) One for the note that gets created when the Support Request is created: https://github.com/MidwestAccessCoalition/lockbox_rails/blob/c534a5b347b82a4ea84b15e298452687e7151e76/lib/create_support_request.rb#L69
https://github.com/MidwestAccessCoalition/lockbox_rails/blob/c534a5b347b82a4ea84b15e298452687e7151e76/app/models/support_request.rb#L71-L77
https://github.com/MidwestAccessCoalition/lockbox_rails/blob/c534a5b347b82a4ea84b15e298452687e7151e76/app/models/note.rb#L6
2) One for the creation of the Support Request itself: https://github.com/MidwestAccessCoalition/lockbox_rails/blob/c534a5b347b82a4ea84b15e298452687e7151e76/lib/create_support_request.rb#L70
3) One for the low balance alert: https://github.com/MidwestAccessCoalition/lockbox_rails/blob/c534a5b347b82a4ea84b15e298452687e7151e76/lib/create_support_request.rb#L71

My change introduced number 1.  Numbers 2-3 were happening previously.  Numbers 1-2 are essentially dupes.

arches: ...which leaves us with three options:

1) leave it
2) don't create a note when the SR is created
3) don't send the specific SR creation_alert

Number 1 is probably kind of annoying, and honestly makes me wish I'd surfaced this question directly - but if we were to choose it, it'd be purely for our own convenience in service of a ship date.  Pros: no changes required.  Cons: double email.

Justification for 2: why do we need a note to record creation, we already have all the information about the SR on the SR itself?  Pros: stop duplicating data in the DB.  Cons: if notes are an audit trail, we're breaking the abstraction a bit.

Justification for 3: no reason to do a one-off when we have a generalized solution for notifications (i.e. notes = notification).  Pros: audit trail remains consistent.  Cons: it becomes more challenging to differentiate SR-creation emails (arguably the most important type) from SR-update emails (perhaps not that far behind creation emails in importance actually).

Gemfile
arches: :+1: sounds good.  If you change your mind I'm still fully happy to pull it into a dev dependency.  Also happy to give a quick tutorial to whoever is going to be supporting prod.

spec/lib/create_support_request_spec.rb
BintLopez: 1. I think this would be really confusing for end users, so let's not do that.

2. I favor this route. I think it's reasonable for audit trail notes to not all result in an email notification. Also support request creation is a very specific action that warrants its own specific messaging. For example, the urgency flag of a support request is prepended to the notification email subject line to convey urgent matters at a glance to lockbox partners receiving an incoming request.

3. If we were to proceed with this route, we'd need to customize the types of messaging for different types of notes. It would be very confusing to a lockbox partner to not be able to differentiate between updates to a case vs incoming requests.

BintLopez: I'm also happy to approve this for merging and then address this comment in a subsequent PR.

BintLopez: I think let's be explicit in this test as to the number of emails expected. I'd like this test to fail if the support request creation action generates more emails.

arches: > support request creation is a very specific action that warrants its own specific messaging

90% agree in theory but 100% agree in practice :)

> I think it's reasonable for audit trail notes to not all result in an email notification

I disagree, and actually this wasn't one of the options I listed (though it could be a fourth).  The problem with special treatment for certain notes is that there isn't anything to differentiate them.  If we want to go down that road (which seems inevitable but not necessarily imminent) we'll need some additional columns on Note.  E.g. we _do_ treat system notes vs manual notes differently, but that's possible because of the `user_id` column.

My option #2 was that we would no longer generate a support-request-creation note, but we would still send a specific support-request-creation email.

BintLopez: Going to go ahead and merge this and transition this convo/decision to an issue.

Issue #298, opened 2019-12-22
Adjust email notifications


jel535s:
## Context
Follow up to https://github.com/MidwestAccessCoalition/lockbox_rails/issues/212, https://github.com/MidwestAccessCoalition/lockbox_rails/pull/288, and https://github.com/MidwestAccessCoalition/lockbox_rails/issues/295

This issue arose out of a discussion on the PR adding an email notification for support request updates:  https://github.com/MidwestAccessCoalition/lockbox_rails/pull/288/files#r355219911

Currently, creation of a new support request will trigger duplicative (but not identical emails)...

<img width="711" alt="image" src="https://user-images.githubusercontent.com/5728859/71278353-f97d0d00-231c-11ea-8250-dfbbdce0c564.png">

<img width="694" alt="image" src="https://user-images.githubusercontent.com/5728859/71278381-0d287380-231d-11ea-90ff-8517bb54d6b4.png">

### Chosen solution:

Option 3 -- leave a note but don't trigger a notification for support request creation
pros: flexibility for different 'types' of notes (logging vs alerts)
cons: notes used as both audit trail and comms between app users

## Process
As a MAC user, create a Support Request.

## Acceptance Criteria/Definition of Done
When a Support Request is created, a system note is created as visible to users in the app.  However, that system-generated note does not trigger an additional email.  Only the "Lockbox Request created" email is sent (not duplicative emails).

Also, we need to adjust the text of the Lockbox Request created email to remove the specific breakdown of transactions. The email should communicate the total amount for the request but does not need to contain the breakdown of spending categories.

## Resources
n/a

PR #315, opened 2020-01-26
Configurable note emails, aka prevent duplication SR creation notifications

arches:
Centralize email alerts within the NoteMailer

## Changelog
- Add `note.notable_action`, with available actions of `create`, `update`, and `annotate`
- Create different email alert templates for different types of notes
- Move the "support request creation" alert template from the SupportRequest mailer to the NoteMailer
- Deprecate unused SupportRequestMailer
- Remove breakdown from support request creation email

## Link to issue:  
Fixes issue #298

n/a

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue


app/mailers/note_mailer.rb
```diff
@@ -1,21 +1,45 @@
 class NoteMailer < ApplicationMailer
 
+  def self.deliver_note_creation_alerts(note)
+    recipients(note).collect do |email|
```
kyaroch: Very minor and non-blocking, but:

- Styleguides usually recommend `#map` over `#collect`, because it's more widely used and thus more recognizable
- Any reason not to use `#each`?

arches: > Any reason not to use #each?

Nope!  Probably a vestige of an earlier approach.

Issue #224, opened 2019-09-29
Application hardening


BintLopez:
## Context
After we are feature complete, we will need to go through a round of application hardening tests and upgrades. This will include things like checking for vulnerabilities in our app with respect to...
- vulnerable dependencies that are out of date
- Cross-Site Request Forgery (CSRF)
- Cross-Site Scripting XSS
- adding `bundle audit`, `snyk`, and/or `brakeman` to our automated deploys so that we'll have a failure and prevent deploys in case of vulnerable dependencies
- Adding attack specs a la https://github.com/DCAFEngineering/dcaf_case_management/blob/master/test/middleware/rack-attack_spec.rb
- Force ssl

## Acceptance Criteria/Definition of Done
Guard against...
- [x] CSRF countermeasures - covered with Rails defaults https://guides.rubyonrails.org/security.html#csrf-countermeasures
- [x] XSS - Add CSP to guard against XSS https://guides.rubyonrails.org/security.html#cross-site-scripting-xss
- [x] Session Fixation - Devise handles this for us by resetting the session after a successful login - https://guides.rubyonrails.org/security.html#session-fixation-countermeasures
- [x] Expire Sessions - handled in Devise; set to expire after 30 mins of inactivity https://github.com/MidwestAccessCoalition/lockbox_rails/blob/master/config/initializers/devise.rb#L176-L179
- [x] Enforce SSL
- [x] Rotate secrets
- [x] Bundle audit and brakeman
- [x] Data encryption at rest
- [x] Filter sensitive info out of logs - https://github.com/MidwestAccessCoalition/lockbox_rails/blob/master/config/initializers/filter_parameter_logging.rb
  - password_confirmation, current_password need to be added to password
  - should we add name_or_alias, email, and name as well?

## Resources
- https://guides.rubyonrails.org/security.html
- OWASP top 10 proactive controls https://owasp.org/www-project-proactive-controls/
- OWASP top 10 risks https://owasp.org/www-project-top-ten/
- OWASP application security verification standard https://owasp.org/www-project-application-security-verification-standard/

PR #321, opened 2020-01-30
Add CSP

BintLopez:
## Changelog
- Adds a content security policy via the secure_headers gem

## Link to issue:  
Part of XSS countermeasures for https://github.com/MidwestAccessCoalition/lockbox_rails/issues/224

- deploying to staging to QA

## Are you ready for review?:

- [ ] Added relevant tests
- [ ] Linked PR to the issue


config/initializers/content_security_policy.rb
```diff
@@ -1,5 +1,54 @@
...
+  config.x_download_options = "noopen"
+  config.x_permitted_cross_domain_policies = "none"
+  config.referrer_policy = %w(origin-when-cross-origin strict-origin-when-cross-origin)
+  config.csp = {
+    # "meta" values. these will shape the header, but the values are not included in the header.
+    preserve_schemes: true, # default: false. Schemes are removed from host sources to save bytes and discourage mixed content.
+    disable_nonce_backwards_compatibility: true, # default: false. If false, `unsafe-inline` will be added automatically when using nonces. If true, it won't. See #403 for why you'd want this.
+
+    # directive values: these values will directly translate into source directives
+    default_src: %w('self'),
```
kyaroch: This was set to `'none'` by default. That setting causes Turbolinks to raise errors sometimes, which is strange, because (a) the default config is supposed to support Turbolinks and (b) the error message refers to default-src itself even though it's supposed to set a default for other settings and not be referenced directly.

However, I can't think of any reason _not_ to do this. Would we ever want to disallow `'self'` for anything? I don't think so.

config/initializers/content_security_policy.rb
```diff
@@ -1,5 +1,54 @@
...
+    child_src: %w('self'), # if child-src isn't supported, the value for frame-src will be set.
+    font_src: %w('self' data: https://fonts.gstatic.com),
+    form_action: %w('self'),
+    frame_ancestors: %w('none'),
+    img_src: %w('self' https://*.amazonaws.com), # Whitelist amazonaws to support the Sqreen image
+    manifest_src: %w('self'),
+    media_src: %w('self'),
+    object_src: %w('self'),
+    sandbox: false, # true and [] will set a maximally restrictive setting
+    script_src: %w('self' 'unsafe-inline'), # unsafe-inline is needed for rails-ujs
```
kyaroch: Not ideal; we may want to create a followup ticket to see if we can turn off `'unsafe-inline'`

config/initializers/content_security_policy.rb
```diff
@@ -1,5 +1,54 @@
...
+    default_src: %w('self'),
+    base_uri: %w('self'),
+    block_all_mixed_content: true, # see http://www.w3.org/TR/mixed-content/
+    child_src: %w('self'), # if child-src isn't supported, the value for frame-src will be set.
+    font_src: %w('self' data: https://fonts.gstatic.com),
+    form_action: %w('self'),
+    frame_ancestors: %w('none'),
+    img_src: %w('self' https://*.amazonaws.com), # Whitelist amazonaws to support the Sqreen image
+    manifest_src: %w('self'),
+    media_src: %w('self'),
```
kyaroch: media_src was `utoob.com` when I picked up this PR. @BintLopez, was that just an example for testing purposes?

config/initializers/content_security_policy.rb
```diff
@@ -1,5 +1,54 @@
...
+    object_src: %w('self'),
+    sandbox: false, # true and [] will set a maximally restrictive setting
+    script_src: %w('self' 'unsafe-inline'), # unsafe-inline is needed for rails-ujs
+    style_src: %w('self' 'unsafe-inline' https://fonts.googleapis.com),
+    worker_src: %w('self'),
+    upgrade_insecure_requests: Rails.env.production?, # see https://www.w3.org/TR/upgrade-insecure-requests/
```
kyaroch: upgrade_insecure_requests breaks local dev, but we absolutely want it on in production. Any thoughts on a better way to do this? We could also add an env var that disables this, although that would create more configuration overhead.

kyaroch:
@BintLopez, I can't request your review on this because you opened the PR, but please review anyway.

config/initializers/content_security_policy.rb
```diff
@@ -1,5 +1,54 @@
...
+# https://www.rubydoc.info/gems/secure_headers/6.3.0#Default_values
+
+SecureHeaders::Configuration.default do |config|
+  config.cookies = {
+    secure: true, # mark all cookies as "Secure"
+    httponly: true, # mark all cookies as "HttpOnly"
+    samesite: {
+      lax: true # mark all cookies as SameSite=lax
+    }
+  }  # Add "; preload" and submit the site to hstspreload.org for best protection.
```
BintLopez, line 16: is this a todo?

config/initializers/content_security_policy.rb
```diff
@@ -1,5 +1,54 @@
...
+  config.x_frame_options = "DENY"
+  config.x_content_type_options = "nosniff"
+  config.x_xss_protection = "1; mode=block"
+  config.x_download_options = "noopen"
+  config.x_permitted_cross_domain_policies = "none"
+  config.referrer_policy = %w(origin-when-cross-origin strict-origin-when-cross-origin)
+  config.csp = {
+    # "meta" values. these will shape the header, but the values are not included in the header.
+    preserve_schemes: true, # default: false. Schemes are removed from host sources to save bytes and discourage mixed content.
+    disable_nonce_backwards_compatibility: true, # default: false. If false, `unsafe-inline` will be added automatically when using nonces. If true, it won't. See #403 for why you'd want this.
```
BintLopez, line 28: What is this #403 reference?

config/initializers/content_security_policy.rb
kyaroch: No, this is a suggestion from the authors of the secure_headers gem.

config/initializers/content_security_policy.rb
BintLopez: lol yeah that was most def an example

config/initializers/content_security_policy.rb
kyaroch: I assume this refers to https://github.com/github/secure_headers/issues/403.

config/initializers/content_security_policy.rb
BintLopez: Is the commented out code on line 71 of this file what we'd need to do to do that?

config/initializers/content_security_policy.rb
BintLopez: Hmmm.... agreed that env var will be more overhead but might also be safer in case we deploy this and things are breaking left and right in prod.

config/initializers/content_security_policy.rb
kyaroch: Good question; let me check.

config/initializers/content_security_policy.rb
kyaroch: I made that suggestion mainly because it feels slightly dirty to me to reference a class method in a configuration file like this. I don't think the possibility that it might break prod is a good reason to do it. `upgrade_insecure_requests` is so critical, IMO, that it might be better to have the site down in prod than to have it turned off in prod.

config/initializers/content_security_policy.rb
kyaroch: Okay, it turns out that line of code won't work with secure_headers, only with the default Rails config. I've spent a few minutes looking at the documentation and I think we need to turn on nonce generation in another way. In my opinion, this should be a followup; I think we should get this out ASAP, and doing so will still make our app more secure than it currently is.

config/initializers/content_security_policy.rb
scottyschup: I'm guessing that's a secure_headers issue or PR

config/initializers/content_security_policy.rb
scottyschup: Definitely. And a terrible one to use as an example since the actual site is one that tries to get you to install browser extensions and does who knows what else shady crap.

config/initializers/content_security_policy.rb
scottyschup: We could define it one way here and override it in the relevant `config/environments` file. Probably set it to `true` here as that's the safer option, and then in `config/environments/development.rb`:
```
::SecureHeaders::Configuration.default do |config|
  config.csp[:upgrade_insecure_requests] = false
end
```
(or something like that)

config/initializers/content_security_policy.rb
```diff
@@ -1,5 +1,54 @@
...
+    media_src: %w('self'),
+    object_src: %w('self'),
+    sandbox: false, # true and [] will set a maximally restrictive setting
+    script_src: %w('self' 'unsafe-inline'), # unsafe-inline is needed for rails-ujs
+    style_src: %w('self' 'unsafe-inline' https://fonts.googleapis.com),
+    worker_src: %w('self'),
+    upgrade_insecure_requests: Rails.env.production?, # see https://www.w3.org/TR/upgrade-insecure-requests/
+  }  # This is available only from 3.5.0; use the `report_only: true` setting for 3.4.1 and below.
+end
+# Below is the Rails auto-generated CSP config in case
```
BintLopez, line 49: Since we're using SecureHeaders gem, I think we can go ahead and delete the comments lines 49 and on...

config/initializers/content_security_policy.rb
BintLopez: Nonce generation as a followup makes sense to me :D Can you make that follow up issue pretty please? 🍨

config/initializers/content_security_policy.rb
BintLopez: Environment based seems worth exploring! I think this can be a follow up with the nonce generation

config/initializers/content_security_policy.rb
BintLopez: for the follow up :D

BintLopez:
Doing a short term fix but logging the issue for folllow ups
image

BintLopez:
https://github.com/MidwestAccessCoalition/lockbox_rails/pull/415

Issue #337, opened 2020-02-02
Any user can edit any other user's notes text


BintLopez:
## Context
Any users that have access to the same support requests can edit each other's notes.

## Process
As a logged in (admin or partner) user, navigate to a support request where multiple users have left multiple notes. Edit a note that's not yours 😳 

## Acceptance Criteria/Definition of Done
A user should only be able to edit their own notes

PR #338, opened 2020-02-02
Ensure only the user that created a note may edit it

misternu:
## Changelog
Made it so that users are the only ones who can edit their own notes. I'm not sure this is the most desired behavior, but here's the PR. It would be easy enough to change again later, perhaps only admins can edit other people's notes, we'll go over this at today's meeting for a bit

## Link to issue:  
#337 

Screen Shot 2020-02-02 at 1 00 03 PM


misternu:
marked dnm needs discussion

misternu:
Discussed this and removed dnm

Issue #332, opened 2020-02-01
Urgency flag displaying empty string


BintLopez:
## Context
Currently support requests are displaying the urgency flag when it's an empty string
![image](https://user-images.githubusercontent.com/5728859/73596365-23624b80-44e7-11ea-8f46-44a4c611fc68.png)

## Process
Create a support request, leave the urgency flag blank, observe that the empty "Urgency Flag" appears

## Acceptance Criteria/Definition of Done
blank urgency flag should not display.

PR #340, opened 2020-02-02
Show urgency flag only if not nil or empty string

misternu:
## Changelog
- Empty string urgency flags do not appear on support requests

## Link to issue:  
#332


app/views/lockbox_partners/support_requests/_flag.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if @support_request.urgency_flag %>
+<% if @support_request.urgency_flag.present? %>
```
BintLopez: :squid:

PR #342, opened 2020-02-02
Disable registration 319

### LINKED ISSUE #319: Disable new user registration & handle nil user role
## Context
Currently, it's possible for any user to register for the app at sign_up, log in, and see all lockbox partners. Sign up should be disabled and only admin users should be able to see all lockbox partners

image

## Process
- navigate to users/sign_up
- follow the sign in steps
- update the confirmed_at field on the user to Time.current on the backend or confirm via the email
- observe that you can log in and see a list of 

- Sign Up should not be an available action
- If a user has no role set, they shouldn't be able to see anything.

## Resources

kyaroch:

## Changelog
- Ensured that `#new`, `#create` and `#destroy` actions in `Users::RegistrationsController` wouldn't work
- Required all users to have a valid role
- Fixed bug that caused users with null role (no longer possible, anyway) to see the admin dash

## Link to issue:  
Fixes #319 

- N/A

## Relevant Screenshots: 
- N/A, as this prevents these views from being rendered

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue


app/models/user.rb
```diff
@@ -8,20 +8,22 @@ class User < ApplicationRecord
          :confirmable, :lockable, :trackable,
          :timeoutable
 
-  scope :confirmed, -> { where.not(confirmed_at: nil) }
```
kyaroch: I moved this to keep the scopes together

Issue #357, opened 2020-02-11
Bug with nil existing_user when resetting password


BintLopez:
## Context
When a new user was attempting to complete the onboarding flow, they hit the following error page when attempting to reset their password...

What the user sees...
![image](https://user-images.githubusercontent.com/5728859/74204852-8025da80-4c3a-11ea-8a1a-b009a0f076a9.png)

Screenshot of server logs...
![image](https://user-images.githubusercontent.com/5728859/74205402-d005a100-4c3c-11ea-8b73-2172b13316d5.png)

## Process
This appears to be due to a stale password reset token. To reproduce...
- do a password reset two times in a row
- click the password reset link in the first email
- the password reset page errors out

This is exaggerated in production due to Mailgun sending delays.

## Acceptance Criteria/Definition of Done
- Stale reset tokens should error out gracefully
  - graceful failure here would display a message to the user that their reset token is invalid and that they should check their email for a more recent password token. If it doesn't show up in 5 minutes, then the issue should be escalated to the lockbox support email addy
- Automated tests should be added for this scenario
- Should change production logging to warn instead of debug to prevent emails w/ reset tokens being leaked in the logs https://github.com/heartcombo/devise#password-reset-tokens-and-rails-logs

PR #359, opened 2020-02-11
render 404 and give flash alert for expired reset token

BintLopez:
## Changelog
- Adds error handling for stale reset token

## Link to issue:  
Fixes issue https://github.com/MidwestAccessCoalition/lockbox_rails/issues/357

- Adding automated tests

## Relevant Screenshots: 
image

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -24,7 +24,7 @@ class Users::PasswordsController < Devise::PasswordsController
     set_existing_user
     super do |resource|
       if resource.errors.empty?
-        resource.update(update_password_params)
+        resource.update(include_name_params)
       end
     end
   end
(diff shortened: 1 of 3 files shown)


misternu:
@BintLopez should be gtg

app/controllers/users/passwords_controller.rb
```diff
@@ -69,9 +68,19 @@ def set_existing_user
...
-    token = Devise.token_generator.digest(
-      self, :reset_password_token, params[:reset_password_token]
-    )
-    @existing_user = User.find_by(reset_password_token: token)
+    @existing_user = User.with_reset_password_token(reset_password_token)
+    record_not_found unless @existing_user && @existing_user.reset_password_period_valid?
+  end
+
+  def reset_password_token
+    params[:reset_password_token] || params[:user][:reset_password_token]
```
BintLopez: When do we expect the reset token to be under `params` vs `params[:user]`?

misternu: when the user clicks the link in their email eg
`<a href="http://localhost:3000/users/password/edit?reset_password_token=ye5smKv3z46yYyzCMekz">Change my password</a>`

misternu: but in the edit form it's nested, and the devise controller seems to expect it these two ways

PR #360, opened 2020-02-16
Display times in the current user's time zone

### LINKED ISSUE #344: Mask timezones with user's local time zones
## Context
We need to display timestamps in the correct timezone -- the app TZ is UTC but users need to see a converted version of that to avoid confusion.  Will this require manual entry at account creation of the user's time zone?  Or can we adjust automatically?

This will probably be linked to and may be combined with #313 

image

## Process
As a logged in user, look at note timestamps.  This should be a global display setting of CST, with the database remaining in UTC.  
Add a user setting in the nav bar where they can change their time zone.
Add a user setting in the onboarding screen when a user clicks on their confirmation email link, where they can set their time zone as well.

As a logged in user, I should see timestamps in my own timezone
However, the database needs to remain in UTC

kyaroch:

## Changelog
- Allow users to set timezone during new user flow and when editing account details
- Display all times in the current user's time zone if it is set

## Link to issue:  
Fixes #344 

- Before casting `RECONCILIATION_INTERVAL.days.ago` to a date in `LockboxPartner#reconciliation_needed?`, I saw test failures when the date was different in CST and UTC. I used Timecop to confirm that this no longer happens.

## Relevant Screenshots: 
Screenshot_20200223_165901
Screenshot_20200223_165937

I am unable to take a screenshot of the open dropdown on my Linux machine, since opening the screenshot utility causes it to lose focus.

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue


config/application.rb
```diff
@@ -13,6 +13,8 @@ class Application < Rails::Application
 
     config.exceptions_app = routes
 
+    config.time_zone = 'Central Time (US & Canada)'
```
kyaroch: Note that this does _not_ affect what is stored in the database:
```
lockbox_rails_development=> SELECT * FROM notes WHERE id=63;
 id |       text        | notable_id |  notable_type  |         created_at         |         updated_at         | user_id | notable_action 
----+-------------------+------------+----------------+----------------------------+----------------------------+---------+----------------
 63 | aaaaaqasfefsdefsa |          3 | SupportRequest | 2020-02-16 22:41:36.764054 | 2020-02-16 22:41:36.764054 |       1 | annotate
```
But it does affect what is displayed in the rails console:
```
[4] pry(main)> Note.last
  Note Load (0.5ms)  SELECT "notes".* FROM "notes" ORDER BY "notes"."id" DESC LIMIT $1  [["LIMIT", 1]]
=> #<Note:0x00007fb9e4463f48
 id: 63,
 text: [FILTERED],
 notable_id: 3,
 notable_type: "SupportRequest",
 created_at: Sun, 16 Feb 2020 16:41:36 CST -06:00,
 updated_at: Sun, 16 Feb 2020 16:41:36 CST -06:00,
 user_id: 1,
 notable_action: "annotate">
```

app/controllers/application_controller.rb
```diff
@@ -1,6 +1,7 @@
 class ApplicationController < ActionController::Base
   before_action :authenticate_user!
   before_action :set_paper_trail_whodunnit
+  around_action :set_time_zone, if: -> { current_user&.time_zone }
```
kyaroch: https://thoughtbot.com/blog/its-about-time-zones

lib/update_support_request.rb
```diff
@@ -58,12 +58,10 @@ def cache_original_values
   def notate_changes
     note_text = []
 
-    timestamp = Time.now.strftime("%Y-%m-%d at %-I:%M%P")
```
kyaroch: Removing this because:
* If we write the time into a string literal, there's no way to display it in the user's time zone
* It's currently redundant with the `Note#created_at` timestamp

app/models/lockbox_partner.rb
```diff
@@ -70,7 +70,9 @@ def historical_actions
   def reconciliation_needed?
     return false unless persisted?
     return false unless !!reconciliation_interval_start
-    reconciliation_interval_start <= RECONCILIATION_INTERVAL.days.ago
+    # Cast the DateTime to a Date, since comparing a Date with a DateTime can
+    # cause unexpected results when the date is different in CST and UTC
+    reconciliation_interval_start <= RECONCILIATION_INTERVAL.days.ago.to_date
```
BintLopez: very helpful comment! muchas gracias! ✨ 🔮 🦄

app/models/support_request.rb
```diff
@@ -73,10 +73,7 @@ def status_options(include_current: false)
   end
 
   def record_creation
-    note_text =
-      "Support request generated by #{user.name} " \
-      "at #{created_at.strftime("%I:%M %p %:::z")} " \
-      "on #{created_at.strftime("%B %d, %Y")}"
```
BintLopez: Can we keep this in or do we lose the timestamp here?

config/application.rb
BintLopez: Discussed in person, and decided that per best practices stated in the Thoughtbot arcticle about timezones, we should _not_ set the timezone at the application level. However, we should set the timezone at the lockbox partner level. That way all accounting events inherit their timezone from the lockbox partner.

app/models/support_request.rb
BintLopez: discussed and decided that this is better represented by the note's created at timestamp and displayed there.

Issue #152, opened 2019-07-28
New lockbox partner initialization flow- ready to implement


BintLopez:
This task encompasses designing and implementing the flow for adding a new lockbox partner as an admin users.

As an admin user, I should be prompted to do the following after filling out the new partner form:
- add users to the lockbox
- add cash to the lockbox

Special cases:
- When a lockbox has no users yet, the add cash form should be disabled.

jel535s:
The second half of this ticket is handled under #52

gujacq:
Here's a link to the lucidchart: https://www.lucidchart.com/invitations/accept/01a9497a-eb2c-4c6d-a2f9-69271ecb9757

Attaching a PDF as well.

Purple = MAC coordinator view
Blue = partner clinic view
Dotted outline = action happening outside of the app

[New lockbox partner initialization flow.pdf](https://github.com/MidwestAccessCoalition/lockbox_rails/files/3563987/New.lockbox.partner.initialization.flow.pdf)

jel535s:
@gujacq - this flow looks great, approved!  Can you make a couple mockups of the messaging for:
- Partner sees the following confirmation message on dashboard: Welcome message, any actionable items needed, "action needed" and "all activity" tab access
- Coordinator app gets alert that partner clinic received cash.  Message should be: "[Lockbox Partner] has confirmed their cash is received, and you can now send them Support Requests."
- "File a support request" tab is live after confirmation (and one where it isn't live).  
- Partner receives in-app notification of money added.  Should reuse the "confirm deposit" message.
- After 3 days of inaction, app pings MAC coordinator to check on clinic.  This should be a red constant bar which appears at the top of the "All Activity" page which is accessed when a MAC user clicks on "View Lockbox Partner Details".  ALSO:
- For UX, this message should also appear on the main page, which is where they go to send Support Requests.  Message should be: "[Lockbox Partner] is not active, has not confirmed cash for X days."
(note for future:  once dropdown for clinics clarifies whether they are active, this last messaging can be redacted from the app)

gujacq:
Partner sees the following confirmation message on dashboard: Welcome message, any actionable items needed, "action needed" and "all activity" tab access: https://invis.io/9EMYOUG7ZBA#/367770324_dashboard_-_Partner_User_-account_Creation_Confirmation-

[...]

PR #372, opened 2020-02-23
152 admin alerts cash confirmation

mrbberra:
## Changelog
- Added alerts for admin users when a cash addition has been pending for > 3 days
- Added alerts for admin users when a lockbox partner has recently (last 48 hrs + no support requests have been sent yet) confirmed their first cash addition
- Refactored some tests (they were incorrectly creating LockboxActions with 'client_support' rather than 'support_client' for the action type)

## Link to issue:  
Addresses issue #152 

- We discussed the possibility of using something more sophisticated in the future than last updated_at to determine whether an initial cash addition was 'recently' confirmed. 

## Relevant Screenshots:
Screen Shot 2020-02-23 at 11 55 51 AM
Screen Shot 2020-02-23 at 11 56 06 AM
Screen Shot 2020-02-23 at 11 56 29 AM
Screen Shot 2020-02-23 at 11 56 38 AM

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue


app/models/lockbox_partner.rb
```diff
@@ -48,6 +48,27 @@ def cash_addition_confirmation_pending?
...
   end
 
+  def longstanding_pending_cash_addition?
+    pending_cash_addition_age >= 3
+  end
+
+  def pending_cash_addition_age
+    earliest_pending_cash_addition = lockbox_actions.pending_cash_additions.order(:eff_date).first
+    return 0 unless earliest_pending_cash_addition
+    (Date.today - earliest_pending_cash_addition.eff_date).to_i
```
BintLopez: We should use `Date.current` here to inherit application timezone config.

app/models/lockbox_partner.rb
```diff
@@ -48,6 +48,27 @@ def cash_addition_confirmation_pending?
...
+  end
+
+  def recently_completed_first_cash_addition?
+    completed_additions = lockbox_actions.completed_cash_additions
+    return false if completed_additions.none?
+    puts lockbox_actions.where(action_type: LockboxAction::SUPPORT_CLIENT).inspect
```
BintLopez: Let's remove this puts statement

app/models/lockbox_partner.rb
```diff
@@ -48,6 +48,27 @@ def cash_addition_confirmation_pending?
...
+    return false if lockbox_actions.where(action_type: LockboxAction::SUPPORT_CLIENT).any?
+    first_cash_addition_completed_at = completed_additions.order(:updated_at).first.updated_at
+    hours_since_first_cash_addition_completed = (Time.now - first_cash_addition_completed_at) / 1.hour
```
BintLopez: Use `Time.current` instead of `Time.now` to get the rails tz config

spec/models/lockbox_partner_spec.rb
```diff
@@ -53,8 +59,72 @@ def pending_request_on(partner, date, amount_breakdown)
     end
   end
 
+  describe '#longstanding_pending_cash_addition?' do
+    let(:partner) { FactoryBot.create(:lockbox_partner) }
+
+    context 'with a longstanding pending cash addition' do
+      let!(:cash_addition) { add_cash(partner, 4.days.ago) }
```
BintLopez: minor thing -- can we use `3.days.ago` here so that we test along the boundary of behavior?

app/models/lockbox_partner.rb
kyaroch: A helpful link I had to dig up while working on my time zone PR: https://thoughtbot.com/blog/its-about-time-zones

(ctrl+F "do's and don'ts" for a list of methods to avoid, although I think `Date.today` is the only one you used)

app/models/lockbox_partner.rb
```diff
@@ -48,6 +48,27 @@ def cash_addition_confirmation_pending?
     lockbox_actions.pending_cash_additions.any?
   end
 
+  def longstanding_pending_cash_addition?
+    pending_cash_addition_age >= 3
```
kyaroch: I'd suggest moving `3` into a constant.

spec/models/lockbox_partner_spec.rb
```diff
@@ -53,8 +59,72 @@ def pending_request_on(partner, date, amount_breakdown)
...
+
+    context 'with a pending cash addition' do
+      let!(:cash_addition) { add_cash(partner, 2.days.ago) }
+      it 'counts the correct number of days' do
+        expect(partner.pending_cash_addition_age).to eq(2)
+      end
+    end
+  end
+
+  describe '#recently_completed_first_cash_addition?' do
```
BintLopez: I think let's get one last test in here that tests that an add cash action completed 49 hours ago will be falsey. We could do this by adding the `Timecop` gem as a test dependency https://rubygems.org/gems/timecop/versions/0.9.1. It was only a matter of time before timecop would be useful for tests, and I feel like once we have it, it will benefit other tests. Cool with doing this as a follow up task too :)

app/models/lockbox_partner.rb
```diff
@@ -48,6 +48,27 @@ def cash_addition_confirmation_pending?
...
+  end
+
+  def recently_completed_first_cash_addition?
+    completed_additions = lockbox_actions.completed_cash_additions
+    return false if completed_additions.none?
+    puts lockbox_actions.where(action_type: LockboxAction::SUPPORT_CLIENT).inspect
+    return false if lockbox_actions.where(action_type: LockboxAction::SUPPORT_CLIENT).any?
+    first_cash_addition_completed_at = completed_additions.order(:updated_at).first.updated_at
+    hours_since_first_cash_addition_completed = (Time.now - first_cash_addition_completed_at) / 1.hour
+    return false unless hours_since_first_cash_addition_completed <= 48
```
kyaroch: Likewise, ideally, pull out `48` into a constant

spec/models/lockbox_partner_spec.rb
```diff
@@ -7,8 +7,8 @@
 
   def add_cash(partner, date)
     LockboxAction.create!(
-      action_type:     'add_cash',
-      status:          'pending',
+      action_type:     LockboxAction::ADD_CASH,
```
kyaroch: :+1:

spec/models/lockbox_partner_spec.rb
kyaroch: Note that https://github.com/MidwestAccessCoalition/lockbox_rails/pull/360 will also add Timecop.

app/models/lockbox_partner.rb
mrbberra: ah shoot my bad

Issue #145, opened 2019-07-21
Add tracking number to add cash


BintLopez:
## Context
As a partner user, I want to be able to see a tracking number for checks being sent to our lockbox, so that I can keep an eye out for them in the mail.

## Process
In the current production instance at https://mac-lockbox.herokuapp.com/ sign in as a MAC user.  Click on "View Lockbox Partner Details" for a lockbox partner, then select the "Add Cash to a lockbox" option in the top bar.  Screenshot under resources for that page.
When submitting a cash add, notice that there is only 1 field, the amount of cash added.

## Acceptance Criteria/Definition of Done
When submitting cash to a lockbox partner as above, we want the form to add one more optional field for a tracking number for checks, or confirmation name for online processing.  Title for field: "(Optional) Tracking number for checks/Confirmation number for online processing"

## Resources
Add cash page from a MAC user perspective:
![image](https://user-images.githubusercontent.com/53353956/74112665-1dfaa600-4b64-11ea-9315-ee94b72ccf5e.png)

AG-Sile:
tech notes: add a new table, tracking_infos connected to lockbox_actions with provider_name and tracking_id

PR #375, opened 2020-02-23
adding tracking_info

AG-Sile:

## Changelog
- What did you do (high level bullet points)?

## Link to issue:  
Fixes issue #145

This adds the tracking info, however, it doesn't currently get displayed anywhere. 
Screen Shot 2020-03-08 at 2 53 10 PM
As a follow up, we should actually display it either by having some sort of details page or include it in the email.

- N/A

## Relevant Screenshots: 

## Are you ready for review?:

- [ ] Added relevant tests
- [ ] Linked PR to the issue


lib/add_cash_to_lockbox.rb
```diff
@@ -20,6 +20,13 @@ def call
         raise ActiveRecord::Rollback
       end
 
+      if tracking_number || delivery_method
```
kyaroch: Should we be checking `#present` here? Could the `tracking_number` or the `delivery_method` be an empty string? If so, this could create a blank `tracking_info`.

BintLopez: good eyes @kyaroch! Yeah... let's use `#present`?

AG-Sile:
@BintLopez done

Issue #374, opened 2020-02-23
Add validations around enum fields


mrbberra:
## Context
There are a number of models that have fields whose value we assume to be one of a short list of possibilities. For example, LockboxActions can have a status of either 'pending', 'completed', or 'cancelled'. 
Mostly creating this issue because I caught the assumption being violated in a test (created a LockboxAction with an action_type of 'client_support' rather than 'support_client'), and it was annoying and seemed potentially dangerous.

## Process
Create a LockboxAction with a type 'client_support'. It ignores all the necessary validations that go along with having the correct type- like needing a related support request.

## Acceptance Criteria/Definition of Done
I'd like to add validations to ensure records actually follow our assumptions.

PR #377, opened 2020-02-23
374 enum validations

mrbberra:
## Changelog
- Adds validations on LockboxActions for status and action_type
- Adds validations on LockboxTransactions for balance_effect and category

## Link to issue:  
Fixes issue #374 

- N/A

## Relevant Screenshots: 
- N/A

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue


app/models/lockbox_action.rb
```diff
@@ -24,7 +24,7 @@ class LockboxAction < ApplicationRecord
     COMPLETED = 'completed',
     CANCELED  = 'canceled'
   ].freeze
-  validates :status, inclusion: STATUSES
+  validates :status, inclusion: STATUSES, allow_nil: true
```
BintLopez: what's the use case for having a nil status?

mrbberra: Honestly I don't know! I included that in order to comply with [existing spec](https://github.com/MidwestAccessCoalition/lockbox_rails/blob/master/spec/models/lockbox_action_spec.rb#L122):
```
    context 'When a lockbox action is created with a nil status' do
      let(:status) { nil }
      it "saves it with a pending status" do
        expect(action).to be_pending
      end
    end
```
I'd be happy to remove this test and the `allow_nil`, but wanted to make sure I wasn't changing existing expected behavior.

BintLopez: hmmmmm... well that is a good reason then ha. Thinking through whether this is a valid use case....

BintLopez: Ahhhhh okay, we had that test because we have a before_save method to set 'pending' as a default status -- https://github.com/MidwestAccessCoalition/lockbox_rails/blob/master/app/models/lockbox_action.rb#L19

The real question is whether we're actually creating any lockbox actions _without_ an explicit status. If we're not, we can get rid of that before_save hook and simply use the validation without allowing nil.

mrbberra: I think we could leave in the hook and just change it to run before_validation instead of before_save? Unless we think the assumption that a nil status should actually be pending is dangerous.

BintLopez: I think that's a fine way to do it! Though we shouldn't be making invalid lockbox_actions in the first place but better handle than error out

spec/models/lockbox_partner_spec.rb
```diff
@@ -22,9 +22,10 @@ def add_cash(partner, date)
 
   def pending_request_on(partner, date, amount_breakdown)
     partner.lockbox_actions.create!(
-      action_type: 'client_support',
-      status:      'pending',
+      action_type: LockboxAction::SUPPORT_CLIENT,
+      status:      LockboxAction::PENDING,
       eff_date:    date,
+      support_request: partner.support_requests.first
```
BintLopez: this is where the merge conflict is... thinking that what we should do here is pass in the support request as an argument so that we explicitly know which SR this action is tied to

mrbberra: addressed by https://github.com/MidwestAccessCoalition/lockbox_rails/pull/377/commits/08b7c2e283663659380722b606ea63690971619d

PR #380, opened 2020-02-23
326 sidekiq workers

misternu:
Adds sideqik and ensures that support status updates are quick

## Link to issue: 
#326

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue

PR #386, opened 2020-03-01
Add redis initializer for heroku

misternu:
Follow up to https://github.com/MidwestAccessCoalition/lockbox_rails/pull/380

### DIFF
--- a/config/initializers/redis.rb
+++ b/config/initializers/redis.rb
@@ -1,0 +1,3 @@
+if ENV["REDISCLOUD_URL"]
+  $redis = Redis.new(:url => ENV["REDISCLOUD_URL"])
+end


config/initializers/redis.rb
```diff
@@ -0,0 +1,3 @@
+if ENV["REDISCLOUD_URL"]
+    $redis = Redis.new(:url => ENV["REDISCLOUD_URL"])
```
BintLopez: this is quite the nitpick, but can we please use 2 spaces for indentation here?

Also... why the `$` variable? I'm not familiar with this....

misternu: $ variables are global, you're not familiar because it's the worst smell for normal development, but the redis instance lives at this variable. https://devcenter.heroku.com/articles/rediscloud#configuring-redis-on-rails

misternu: I can remove the spaces but they're there because this is an exact copy paste from heroku's article

misternu: https://ruby-doc.org/docs/ruby-doc-bundle/UsersGuide/rg/globalvars.html

Issue #284, opened 2019-12-02
V2 Help page/ticketing


jel535s:
## Context
For the 2nd Version of the app, we need to build out the help/support contact page which is accessible by all registered and logged in users.  This page needs to be updated to have a form which will automatically post to github as part of our ticketing process for issues.

## Process
The Help page should be updated to add a form which will automatically post to github as part of our ticketing process for issues.

## Acceptance Criteria/Definition of Done

Step 1- @gujacq - please mock-up a form for the help page.  This form should collect:

- _This copy at the top:_  
For MAC users:
If you're having issues in the app, please check out our FAQ manual here first [LINK: https://docs.google.com/document/d/1GEq8hbnjmH_4xq1WFgvuVrLMSTU1knjHLrKZyVqKzcY/edit?usp=sharing]
If you aren't able to find the answer there, you can also fill out this form and click submit to log your issue.  Our all-volunteer IT team will start work on it right away, but the issue may remain for some time depending on how heavy their workload is this week.  
For Partner users: 
If you're having issues in the app, please check out our FAQ manual here first [LINK: https://docs.google.com/document/d/1bhUsBvOoS8E15s8XPrQMejrZONrp-DuZJ42AjPf2odw/edit?usp=sharing]
If you aren't able to find the answer there, you can also fill out this form and click submit to log your issue.  Our all-volunteer IT team will start work on it right away, but the issue may remain for some time depending on how heavy their workload is this week.  
- user email address (ideally auto-fill)
- date/time (ideally auto-fill)
- Issue (long-form field)
- place to upload screenshots from phone or computer where they are accessing the app
- submit button
- _This copy at the bottom, near the submit button:_  If you need immediate assistance, please email lockbox@midwestaccesscoalition.org.

Step 1- dev team- please implement the copy/mockups as listed below.

## Resources

gujacq:
Support form for partner clinics (though the design should be the same for MAC users, just different content—see Jesse's notes above): https://invis.io/9EMYOUG7ZBA#/404321025_dashboard_-_Partner_User_-get_Support-_2

What it looks like when someone uploads multiple files - https://invis.io/9EMYOUG7ZBA#/404321027_dashboard_-_Partner_User_-get_Support-_2_Screenshots

Confirmation screen - https://invis.io/9EMYOUG7ZBA#/404321026_dashboard_-_Partner_User_-get_Support-_2_Confirmation

nsiwnf:
[...]

PR #421, opened 2020-04-17
Adding link to google form

nsiwnf:
## Changelog
Updated Get Support page to include Google Form link for ticket request

## Link to issue:  
#284 

Originally we were going to embed the form but then wanted to mitigate security issues by using an existing form. We decided to go with Google Forms to give tech people access to the spreadsheets. Google forms cannot be embedded if they include file upload sections so we decided to make it just a link instead of an iframe.

## Relevant Screenshots: 
Screen Shot 2020-05-10 at 2 32 48 PM

## Are you ready for review?:

- [ ] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/app/views/dashboard/support.html.erb
+++ b/app/views/dashboard/support.html.erb
@@ -3,11 +3,22 @@
   <p>
     If you're having issues in the app, please check out
     <% user_type = current_user.admin? ? "ADMIN" : "PARTNER" %>
-    <%= link_to "our FAQ manual", ENV["LOCKBOX_#{user_type}_FAQ_LINK"] %>
-    first.
-  </p>
-  <p>
-    If you can't find the answer there, you can also email
-    <%= link_to ENV["LOCKBOX_EMAIL"], "mailto:#{ENV["LOCKBOX_EMAIL"]}" %>.
-  </p>
-</div>
+      <%= link_to "our FAQ manual", ENV["LOCKBOX_#{user_type}_FAQ_LINK"] %>
+      first.
+    </p>
+    <p>
+      If you aren't able to find the answer there, you can also fill out this form and click submit to log your issue. Our all-volunteer IT team will start work on it right away, but the issue may remain for some time depending on how heavy their workload
+      is this week.
+    </p>
+
+    <p>
+      <a href="https://forms.gle/tw6MJQTgjfZhq2nD7" target="_blank">Fill out form</a> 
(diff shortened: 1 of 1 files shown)


app/views/dashboard/support.html.erb
```diff
@@ -3,11 +3,22 @@
...
-  <p>
-    If you can't find the answer there, you can also email
-    <%= link_to ENV["LOCKBOX_EMAIL"], "mailto:#{ENV["LOCKBOX_EMAIL"]}" %>.
-  </p>
-</div>
+      <%= link_to "our FAQ manual", ENV["LOCKBOX_#{user_type}_FAQ_LINK"] %>
+      first.
+    </p>
+    <p>
+      If you aren't able to find the answer there, you can also fill out this form and click submit to log your issue. Our all-volunteer IT team will start work on it right away, but the issue may remain for some time depending on how heavy their workload
```
BintLopez, line 10: can we hyperlink `fill out this form` to link to the form?

app/views/dashboard/support.html.erb
```diff
@@ -3,11 +3,22 @@
...
+      is this week.
+    </p>
+
+    <p>
+      <a href="https://forms.gle/tw6MJQTgjfZhq2nD7" target="_blank">Fill out form</a>
```
BintLopez, line 15: let's make this a button with "Log an Issue or Feedback" instead of "Fill out form"

BintLopez: Also, didn't catch this last week, but let's also pop this in an env var so that we can easily swap out the form as needed.

config/local_env.yml
```diff
@@ -3,3 +3,4 @@
 'LOCKBOX_EMAIL': "kittens@lockbox-example.com"
 'LOCKBOX_PARTNER_FAQ_LINK': "http://www.example.com/partner_faq"
 'REPLY_TO_EMAIL': "no-reply@lockbox-example.com"
+'ISSUE_FORM_LINK': "https://forms.gle/tw6MJQTgjfZhq2nD7"
```
BintLopez, line 6: let's not put the real form link in the public repo

nsiwnf: I removed it from the repo and squashed the commits so they wouldnt show up in the git history - Is there a way to remove it from these conversations too?

Issue #402, opened 2020-03-15
Display cash addition tracking info


BintLopez:
## Context
- Follow up to https://github.com/MidwestAccessCoalition/lockbox_rails/pull/375
- Need to display the delivery method and tracking information to the lockbox partners in the cash addition alert as well as in the notification email about incoming cash

## Process
- As an admin user, add cash to a lockbox -- note that you can enter in tracking info and delivery method
- As a partner user, log in and see your confirm cash notification -- it does not display any of this tracking info
- Email alerts also do not contain the tracking info

## Acceptance Criteria/Definition of Done
Tracking info/delivery method displays in two places:
1. cash addition alert in partner dashboard
2. cash addition notification email

## Resources

PR #427, opened 2020-04-19
Adding tracking info to alert and email for "Add Cash"

nsiwnf:
## Changelog
- Adding "Tracking Info" to alert and email for "Add Cash" action

## Link to issue:  
#402 

- There weren't any mocks provided in ticket. Please verify that it looks ok
- Checked in prod db and there aren't any conflicts/issues with existing data - there arent any lockbox actions that have multiple tracking infos

## Relevant Screenshots: 
(Email)
Screen Shot 2020-05-17 at 3 30 36 PM
(App alert)
Screen Shot 2020-05-17 at 3 28 50 PM

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/app/models/lockbox_action.rb
+++ b/app/models/lockbox_action.rb
@@ -5,7 +5,7 @@ class LockboxAction < ApplicationRecord
   belongs_to :support_request, optional: true
   has_many :lockbox_transactions, dependent: :destroy
   has_many :notes, as: :notable
-  has_many :tracking_infos
+  has_one :tracking_info
 
   accepts_nested_attributes_for :lockbox_transactions, reject_if: :all_blank,
     allow_destroy: true
@@ -140,6 +140,12 @@ class LockboxAction < ApplicationRecord
 
   def debit?
     lockbox_transactions.first&.balance_effect == LockboxTransaction::DEBIT
+  end
+
+  def tracking_info_formatted
+    if tracking_info.present?
+      "#{tracking_info.delivery_method} #{tracking_info.tracking_number}"
+    end
   end
 
   private
--- a/app/views/lockbox_partners/_alerts.html.erb
+++ b/app/views/lockbox_partners/_alerts.html.erb
@@ -24,8 +24,7 @@
         <div>A $<%= cash_action.amount %> check was mailed on <%= cash_action.eff_date.try(:strftime, "%B %-d") %>, please confirm it was received.</div>
 
         <% if cash_action.tracking_info.present? %>
-          <div>Tracking info:</div>
-          <div><%= cash_action.tracking_info_formatted %></div>
+          <div>Tracking info: <%= cash_action.tracking_info_formatted %></div>
         <% end %>
       </div>
       <div class="p-2">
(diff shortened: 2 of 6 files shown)


app/views/lockbox_partners/_alerts.html.erb
```diff
@@ -22,6 +22,13 @@
       <div class="p-2 flex-grow-1">
         <h4 class="alert-heading">Cash Sent</h4>
         <div>A $<%= cash_action.amount %> check was mailed on <%= cash_action.eff_date.try(:strftime, "%B %-d") %>, please confirm it was received.</div>
+
+        <% if cash_action.tracking_infos.present? %>
+          <div>Tracking info:</div>
+          <% cash_action.tracking_infos.each do |tracking_info| %>
+            <div><%= tracking_info.delivery_method %> <%= tracking_info.tracking_number %></div>
```
BintLopez, line 29: Though this is a `has_many`, it doesn't really make sense for an action to have more than one tracking info. I'd say what we can do is either... only select the last one OR change the relation to a `has_one`. Either way, since we know that this is going to only be one thing, I think we can just display the tracking info in the same line.

I'd suggest moving the tracking infos display logic to the LockboxAction model and then testing that it fetches it properly from there. Then in the mailer and the alerts, it's just calling a method on the lockbox_action.

nsiwnf: Sounds good - do you remember the original reason it was `has_many` vs `has_one`?

BintLopez: Trying to recall... I think maybe was thinking of the case where a check gets lost in transit and we have to void and send another? But that still doesn't fully make sense because that would probs be handled best by canceling the add cash and adding another one with the new info.

app/models/lockbox_action.rb
```diff
@@ -5,7 +5,7 @@ class LockboxAction < ApplicationRecord
   belongs_to :support_request, optional: true
   has_many :lockbox_transactions, dependent: :destroy
   has_many :notes, as: :notable
-  has_many :tracking_infos
+  has_one :tracking_info
```
BintLopez, line 8: 👍

app/views/lockbox_partners/_alerts.html.erb
```diff
@@ -22,6 +22,11 @@
       <div class="p-2 flex-grow-1">
         <h4 class="alert-heading">Cash Sent</h4>
         <div>A $<%= cash_action.amount %> check was mailed on <%= cash_action.eff_date.try(:strftime, "%B %-d") %>, please confirm it was received.</div>
+
+        <% if cash_action.tracking_info.present? %>
+          <div>Tracking info:</div>
+          <div><%= cash_action.tracking_info_formatted %></div>
```
BintLopez, line 28: nitpick... can we combine these two divs so these are on the same line?

app/views/lockbox_partners/_alerts.html.erb
nsiwnf: Ah that makes sense - yeah, I think it would be better as a 1-1 relation

app/models/lockbox_action.rb
BintLopez: Let's add a test for this relation in the lockbox_action_spec

PR #433, opened 2020-04-26
Resolve csp issues 419

### LINKED ISSUE #419: Resolve outstanding issues with CSP
## Context
A content security policy was implemented in https://github.com/MidwestAccessCoalition/lockbox_rails/pull/321. However, some issues remain:

1. Currently, `script-src` includes `unsafe-inline`. This raises security issues, and we want to disable `unsafe-inline`. However, that currently breaks rails-ujs. In order to fix this, we need to enable nonce generation.
2. Some commented-out code config/initializers/content_security_policy.rb is now superfluous and should be removed.
3. We may want to take a different approach to enabling/disabling upgrade-insecure-requests based on environment.

## Process
N/A

What do you want to happen?
Possible fixes?

## Resources
Add screenshots/mocks if relevant
Add relevant labels

kyaroch:
## Changelog
- Remove secure-headers gem and return to built-in Rails CSP tools
- Replace `upgrade-insecure-requests` with `block-all-mixed-content`
- Make a few other directives more restrictive to remove potentially insecure sources we're not usign

## Link to issue:  
Resolves #419 

- The best way to QA this is to test all the functionality with the console open, so you'll see if any resources are being improperly blocked (and errors logged). I did this with every page and form I could think of, and found nothing.

## Relevant Screenshots: 
- N/A

## Are you ready for review?:

- [ ] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -10,17 +10,17 @@
 
 Rails.application.config.content_security_policy do |policy|
   policy.base_uri        :self
-  policy.default_src     :self, :https
-  policy.font_src        :self, :https, 'https://fonts.gstatic.com', 'https://demo-lockbox.herokuapp.com'
+  policy.default_src     :self
+  policy.font_src        :self, 'https://fonts.gstatic.com', 'https://demo-lockbox.herokuapp.com'
   policy.form_action     :self
   policy.frame_ancestors :none
-  policy.img_src         :self, :https, 'https://*.amazonaws.com'
+  policy.img_src         :self, 'https://*.amazonaws.com'
   policy.object_src      :none
-  policy.script_src      :self, :https
-  policy.style_src       :self, :https, :unsafe_inline, 'https://fonts.googleapis.com'
-  policy.worker_src      :self, :https
+  policy.script_src      :self
+  policy.style_src       :self, :unsafe_inline, 'https://fonts.googleapis.com'
+  policy.worker_src      :self
   # If you are using webpack-dev-server then specify webpack-dev-server host
-  policy.connect_src     :self, :https, "http://localhost:3035", "ws://localhost:3035" if Rails.env.development?
+  policy.connect_src     :self, "http://localhost:3035", "ws://localhost:3035" if Rails.env.development?
 
   policy.block_all_mixed_content true
   # Specify URI for violation reports
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -14,7 +14,7 @@ Rails.application.config.content_security_policy do |policy|
   policy.font_src        :self, 'https://fonts.gstatic.com', 'https://demo-lockbox.herokuapp.com'
   policy.form_action     :self
   policy.frame_ancestors :none
-  policy.img_src         :self, 'https://*.amazonaws.com'
+  policy.img_src         :self, 'https://*.amazonaws.com' # Whitelist amazonaws to support the Sqreen image
   policy.object_src      :none
   policy.script_src      :self
   policy.style_src       :self, :unsafe_inline, 'https://fonts.googleapis.com'
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -6,46 +6,46 @@
 # default config documentation is here:
 # https://www.rubydoc.info/gems/secure_headers/6.3.0#Default_values
 
-SecureHeaders::Configuration.default do |config|
-  config.cookies = {
-    secure: true, # mark all cookies as "Secure"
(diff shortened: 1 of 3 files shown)


config/initializers/content_security_policy.rb
```diff
@@ -1,79 +1,39 @@
...
+  policy.frame_ancestors :none
+  policy.img_src         :self, :https, 'https://*.amazonaws.com'
+  policy.object_src      :none
+  policy.script_src      :self, :https
+  policy.style_src       :self, :https, :unsafe_inline, 'https://fonts.googleapis.com'
+  policy.worker_src      :self, :https
+  # If you are using webpack-dev-server then specify webpack-dev-server host
+  policy.connect_src     :self, :https, "http://localhost:3035", "ws://localhost:3035" if Rails.env.development?
+
+  policy.block_all_mixed_content true
```
kyaroch, line 25: We have two options here:
1. `upgrade_insecure_requests`: if a stylesheet, image, etc. has an `http://` URL, the browser will use `https://` instead
2. `block_all_mixed_content`: if the page is loaded using `https://`, and a stylesheet, image, etc. has an `http://` URL, the browser will not load it at all

From what I can find online, there's no benefit to using both. I think `block_all_mixed_content` is preferable because there should be no resources loaded insecurely in prod, and this has the additional benefit that we can use the same config in development and prod (since it will do nothing if the page is loaded insecurely).

This assumes that the prod site is only accessible using HTTPS. That's true, right?

config/initializers/content_security_policy.rb
```diff
@@ -1,79 +1,39 @@
...
-# Below is the Rails auto-generated CSP config in case
-# we want to go back to this from SecureHeaders
-
 # Define an application-wide content security policy
 # For further information see the following documentation
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
 
-# Rails.application.config.content_security_policy do |policy|
-#   policy.default_src :self, :https
-#   policy.font_src    :self, :https, :data
```
kyaroch, line 58: I removed `:data` everywhere because its purpose is to enable data URLs. I don't think we use these, and MDN says they're a potential security risk.

config/initializers/content_security_policy.rb
```diff
@@ -1,79 +1,39 @@
...
+#   the intended values are`the same
+
+Rails.application.config.content_security_policy do |policy|
+  policy.base_uri        :self
+  policy.default_src     :self, :https
+  policy.font_src        :self, :https, 'https://fonts.gstatic.com', 'https://demo-lockbox.herokuapp.com'
+  policy.form_action     :self
+  policy.frame_ancestors :none
+  policy.img_src         :self, :https, 'https://*.amazonaws.com'
+  policy.object_src      :none
```
kyaroch, line 18: Governs `<object>`, `<embed>`, and `<applet>`. I don't think we use these, nor should we.

config/initializers/content_security_policy.rb
BintLopez: This should be true. Because we have force_ssl enabled in prod....
https://github.com/MidwestAccessCoalition/lockbox_rails/blob/master/config/environments/production.rb#L46-L47

config/initializers/content_security_policy.rb
BintLopez: This makes sense. From what I'm reading on mozilla docs, this would also disable <iframe> right? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/object-src

config/initializers/content_security_policy.rb
```diff
@@ -1,79 +1,39 @@
...
+# * media_src and manifest_src, because they will default to default_src, and
+#   the intended values are`the same
+
+Rails.application.config.content_security_policy do |policy|
+  policy.base_uri        :self
+  policy.default_src     :self
+  policy.font_src        :self, 'https://fonts.gstatic.com', 'https://demo-lockbox.herokuapp.com'
+  policy.form_action     :self
+  policy.frame_ancestors :none
+  policy.img_src         :self, 'https://*.amazonaws.com'
```
BintLopez, line 17:
```suggestion
  policy.img_src         :self, 'https://*.amazonaws.com' # Whitelist amazonaws to support the Sqreen image
```

config/initializers/content_security_policy.rb
kyaroch: No, I don't think this disables `<iframe>`. The doc page uses `<iframe>` as a counterexample of a tag that's still receiving new features.

In this config, iframes are currently allowed, but are restricted to `self` because `frame-src` falls back to `default-src`.

Issue #431, opened 2020-04-26
Editing a support request fails


BintLopez:
## Context
Currently, coordinators are unable to edit support requests.

## Process
File a support request
Attempt to edit the support request
Observe the error

## Acceptance Criteria/Definition of Done
- As a coordinator, I should be able to edit a support request
- A test should fail for this issue

## Resources
![image](https://user-images.githubusercontent.com/5728859/80320063-c6bab800-87d9-11ea-9455-c921a3835aef.png)

![image](https://user-images.githubusercontent.com/5728859/80320064-cae6d580-87d9-11ea-9829-840ffabd6bd0.png)

gujacq:
Fixed, but still need to add tests

gujacq:
Sree: wrote tests, it's passing locally, but failing at integration so taking a look at that

PR #434, opened 2020-04-26
Fix - Coordinators unable to add transaction to support request

nsiwnf:
## Changelog
- Fixed alert to render on error
- Fixed issue with not saving when adding transaction

## Link to issue:  
#431 

- N/A

## Relevant Screenshots: 

## Are you ready for review?:

- [ ] Added relevant tests
- [ ] Linked PR to the issue

### DIFF
--- a/app/controllers/lockbox_partners/support_requests_controller.rb
+++ b/app/controllers/lockbox_partners/support_requests_controller.rb
@@ -86,7 +86,12 @@ class LockboxPartners::SupportRequestsController < ApplicationController
       flash[:notice] = "Support request was successfully updated"
       redirect_to lockbox_partner_support_request_path(@support_request)
     else
-      render 'edit'
+      render json: {
+        error: render_to_string(
+          partial: 'shared/error',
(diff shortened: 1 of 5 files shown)


misternu:
https://github.com/MidwestAccessCoalition/lockbox_rails/commit/8ad45cab21076ed7e54efefd5f655feb79636a47 adds a validation that a lockbox transaction must have a debit or credit balance_effect, lockbox_partner/support_requests_controller#update does not flash an error when updatesupportrequest fails and 500s anyway

misternu:
UpdateSupportRequest should be updated to include that all lockbox transactions are debits, and support request controller should be updated to show errors and set the @path variable on failure, I can commit on this

misternu:
Instead of changing update support request, I set balance_effect to debit by default, as the support request edit page isn't going to have balance effect in its ui. It also should now properly show errors on update instead of 500ing when trying to render edit

nsiwnf:
@BintLopez @misternu  I'm good with merging this - still working on an end-to-end test which I can commit in a follow-up PR

PR #437, opened 2020-05-03
fix: adds dollar sign to balance display

rebeldroid12:
## Changelog
Adds dollar sign to balance.
https://apidock.com/rails/ActionView/Helpers/NumberHelper/number_to_currency

## Link to issue:  
Fixes issue #N/A

- N/A

## Relevant Screenshots: 

## Are you ready for review?:

- [ ] Added relevant tests
- [ ] Linked PR to the issue

### DIFF
--- a/app/views/lockbox_partners/_header.html.erb
+++ b/app/views/lockbox_partners/_header.html.erb
@@ -6,7 +6,7 @@
   <h2>
     <%= lockbox_partner.name %>
     <span class="float-right">
-      <%= lockbox_partner.balance %>
+      <%= number_to_currency lockbox_partner.balance %>
     </span>
   </h2>
   <div class="lockbox-partner-info">


app/views/lockbox_partners/_header.html.erb
```diff
@@ -6,7 +6,7 @@
   <h2>
     <%= lockbox_partner.name %>
     <span class="float-right">
-      <%= lockbox_partner.balance %>
+      <%= number_to_currency lockbox_partner.balance %>
```
scottyschup, line 9: Looks good. My only thought is are we sure this is always going to be a valid number? Because if not—if some user input could potentially corrupt the number then we might want to  raise an error when that happens:
```
number_to_currency("123a456") # => $123a456
number_to_currency("123a456", raise: true) # => InvalidNumberError
```
But if that's a generated value and we already have sanity checks upstream, then this is all good as is.

BintLopez: one thing to note here is that balance should be a `money` object rather than a float or int. Will `number_to_currency` handle that appropriately?

Issue #387, opened 2020-03-01
Two notifications need building around reconciliation


jel535s:
## Context
1.  Notification via email is needed for ENV['LOCKBOX_EMAIL'] and ENV['FINANCE_EMAIL'] when a reconciliation is completed.  
2.  Notification for past due reconciliation -- should notify ENV['LOCKBOX_EMAIL'], not ENV['FINANCE_EMAIL']

## Process
1.  When a partner user completes a reconciliation, currently no notification is sent to the MAC side.
2.  Similarly, when reconciliations remain past due, no notifications are sent to the MAC side.

## Acceptance Criteria/Definition of Done
1.  When a partner user completes a reconciliation, an email needs to be sent to ENV['LOCKBOX_EMAIL'] and ENV['FINANCE_EMAIL'].  In that email, the amount of the reconciliation is important info.  Copy should read:  _"COPY NEEDED"_
2.  After 7 days of reconciliation being past due, a notification should be sent to only ENV['LOCKBOX_EMAIL'], not ENV['FINANCE_EMAIL']

gujacq:
Moss should have PR for this today

gujacq:
Could be a CC

mrbberra:
per slack conversation, will be using heroku scheduler to implement this

gujacq:
From Veronica re: How often this should notify:

Sorry - missed this previously somehow. I think a weekly follow up until the monthly reconciliation is completed would be the most effective— can we ensure that they run on a Monday or Tuesday morning (I’m afraid Thurs/Fri / later in the day emails will be more easily ignored).

BintLopez:
@misternu -- I think we're going to need a follow up for this task. The rake task is currently erroring out (locally and when running on prod)

```
be rake scheduled_tasks:overdue_reconciliation_emails
rake aborted!
LocalJumpError: unexpected return
```

PR #438, opened 2020-05-10
387 notifications reconciliation

mrbberra:

## Changelog
- Created notifications around reconciliation, to be sent to MAC staff.
- Notify MAC staff when reconciliation is completed for any lockbox partner
- Notify MAC staff when reconciliation has been overdue for a week for any lockbox partner. Repeat this notification weekly (on Mondays) until reconciliation is complete.

## Link to issue:  
Fixes issue #387 

- Wrote a rake task to be run by a cron job. That cron job still has to be written though.

## Relevant Screenshots: 
- N/A

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue


lib/reconciliation.rb
```diff
@@ -39,6 +39,8 @@ def call
           raise ActiveRecord::Rollback
         end
       end
+
+      LockboxPartnerMailer.with(lockbox_partner: lockbox_partner, amount: amount).reconciliation_completed_alert.deliver_now
```
misternu, line 43: Need to move this mailer call into sidekiq worker, to ensure performance

misternu:
Lumped in some deprecation fixes, rspec-benchmarking. These emails should always be sent by sidekiq workers to ensure we don't get a repeat of the note emails issue.

nsiwnf:
I merged from master and now this test it failing 
```  
1) Reconciliation when the amount is valid is performant
     Failure/Error: expect { reconcile }.to perform_under(0.025).sec
       expected block to perform under 25 ms, but performed above 29.4 ms (± 0 ns) 
     # ./spec/lib/reconciliation_spec.rb:84:in `block (3 levels) in <top (required)>'

```

spec/lib/reconciliation_spec.rb
```diff
@@ -77,5 +77,11 @@ def reconcile
       expect(lockbox_transaction.amount).to eq(Monetize.parse('100.00'))
       expect(lockbox_transaction.balance_effect).to eq(LockboxTransaction::CREDIT)
     end
+
+    it 'is performant' do
+      expect(LockboxPartnerMailer).not_to receive(:with)
+      expect(LockboxPartnerMailer).not_to receive(:reconciliation_completed_alert)
+      expect { reconcile }.to perform_under(0.025).sec
```
nsiwnf, line 84: Whats the significance of 0.025?

misternu:
some of those benchmark tests will flake yes, I plan on taking them out once it's been a while since we had complaints about performance for things that send out mail

lib/reconciliation.rb
```diff
@@ -39,6 +39,8 @@ def call
           raise ActiveRecord::Rollback
         end
       end
+
+      ReconciliationCompletedMailerWorker.perform_async(lockbox_partner.id, amount)
```
BintLopez, line 43: If the active record transaction block has any failures, will this line still run?

spec/lib/reconciliation_spec.rb
BintLopez: Do we have a test of the positive case where the reconciliation completed alert _would_ be sent?

spec/models/lockbox_partner_spec.rb
```diff
@@ -490,6 +490,40 @@ def pending_request_on(partner, date, amount_breakdown, request)
...
+      before { reconciliation_action }
+
+      let(:reconciliation_date) do
+        (LockboxPartner::RECONCILIATION_INTERVAL - 1).days.ago
+      end
+
+      it { is_expected.to be false }
+    end
+
+    context 'when the lockbox was last reconciled long enough ago to necessitate a notification' do
```
BintLopez, line 516: We need one more test case for when outside the reconciliation interval, but not long enough ago to necessitate a notification.

app/models/lockbox_partner.rb
```diff
@@ -114,6 +114,17 @@ def reconciliation_interval_start
     start_date&.to_date
   end
 
+  private
+
+  def reconciliation_over_n_days_ago?(num_days)
+    return false unless persisted?
+    return false unless !!reconciliation_interval_start
```
BintLopez, line 121: Since this is a private method, we should be testing behavior in these two guard clauses in our test coverage of #reconciliation_needed?

lib/reconciliation.rb
misternu: I don't think it will, should we change the requirements so that emails are sent when there's a failure?

spec/models/lockbox_partner_spec.rb
misternu: This spec was not actually testing the described method, looks like it had been copy pasted but not changed, fixed now in new commit

PR #443, opened 2020-05-18
Fix nav bar open/close bug #365

### LINKED ISSUE #365: Header nav menu open/close bug
## Context
There's currently a fun little bug that occurs when you complete a specific series of actions in the header nav area (see below).

## Process
As a Lockbox partner user (e.g. `fluffy@catsclinic.com`) logged in on the home page:
1. open the navbar (click the gear icon)
1. with the navbar still open, click the "All Activity" subnav button (the navbar should close)
1. now click the "Action Needed" subnav button and note:
   * The navbar flickers open and closed
   * Pressing "Action Needed" again will produce the same result a second time
   * _however_, any subsequent clicking on the subnav buttons does not

Without any investigation, I'm guessing some `toggle` function is getting double called—perhaps two event listeners are being fired, in which case adding `Event#stopPropagation` to the higher handler may resolve it. But not positive.

Buggy behavior is squashed. Code is merged to master.

## Resources
header_navbar_flicker_bug_slower

scottyschup:
## Link to issue:  
Fixes #365

## Changelog
Refactor nav bar functionality: 
* remove bootstrap classes/attrs
* add custom JS/CSS to control header nav drawer
* add trap focus to nav drawer
* add tests

- See description of bug in issue #365; use repro instructions to test
- Resolved behavior should look like this:

pr443-issue365-header-nav-bug-resolution

## Are you ready for review?:
Yep.

- [x] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/app/assets/stylesheets/dashboard.scss
+++ b/app/assets/stylesheets/dashboard.scss
@@ -35,7 +35,7 @@
 
     a:hover {
       color: white;
-      border: 1px solid darken($color-primary, 10%);
+      border: 2px solid darken($color-primary, 10%);
       background-color: darken($color-primary, 10%);
     }
 
--- a/app/assets/stylesheets/navbar.scss
+++ b/app/assets/stylesheets/navbar.scss
@@ -52,10 +52,11 @@ header {
     border: 2px solid white;
   }
   #navbar {
+    flex-grow: 1;
     display: none;
 
     &.expanded {
-      display: contents;
+      display: block;
     }
 
     .container {
--- a/app/assets/stylesheets/navbar.scss
+++ b/app/assets/stylesheets/navbar.scss
@@ -3,12 +3,15 @@ header {
   background-color: $color-primary;
   font-family: $font-sans;
   color: white;
+
   #nav-title {
     font-weight: bolder;
     color: white;
   }
+
   .links {
     margin-bottom: 0;
+
     a,
     a:visited {
       color: white;
@@ -16,21 +19,26 @@ header {
       font-size: 0.8em;
     }
   }
+
   button.navbar-control {
     background-color: transparent;
     border: none;
     color: white;
+
     &.expanded .fa-bars,
     &.expanded .fa-cog {
       display: none;
     }
+
     &:not(.expanded) .fa-times {
       display: none;
     }
   }
+
   .container {
     width: 80%;
   }
+
   .btn {
     display: block;
     float: right;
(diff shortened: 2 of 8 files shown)


app/views/layouts/_header.html.erb
```diff
@@ -1,4 +1,4 @@
-<header role="banner">
+<header>
```
scottyschup, line 1: The `<header>` tag and the `banner` role serve the same purpose, so having both is unnecessary.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Banner_role

app/views/layouts/_nav_bar.html.erb
```diff
@@ -1,46 +1,52 @@
-<nav class="navbar navbar-default d-flex" role="navigation">
+<nav class="navbar navbar-default d-flex">
```
scottyschup, line 1: `<nav>` and `<div role="navigation">` are basically identical. But semantic tags are preferred over roles whenever possible.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_role

app/assets/stylesheets/dashboard.scss
```diff
@@ -35,7 +35,7 @@
 
     a:hover {
       color: white;
-      border: 1px solid darken($color-primary, 10%);
+      border: 2px solid darken($color-primary, 10%);
```
scottyschup, line 38: Despite being `<a>` tags, these represent buttons and the normal button border is 2px, so this being set at 1px was causing some jerkiness on hover.

app/assets/stylesheets/navbar.scss
```diff
@@ -1,70 +1,112 @@
 header {
   margin: 0;
   background-color: $color-primary;
-  font-family: $font-sans;
-  color: white;
+  font-family: $font-serif;
```
scottyschup, line 4: This font-family more closely aligns with the InVision.

app/assets/stylesheets/navbar.scss
```diff
@@ -74,31 +116,6 @@ header {
...
-  margin-top: 1em;
-  font-size: 0.75em;
-  padding: 0.25em 2em;
-}
-
 .fa-cog {
   font-size: 1.5em;
 }
-
-@media screen and (max-width: 481px) {
```
scottyschup, line 90: By removing `flex` and `float`s that were not playing well with each other, the two columns of the navbar drawer now just collapse when they need to based on their content rather than at some arbitrary breakpoint making this rule unnecessary.

BintLopez:
Manually QA-ed and this is looking good, but we have a test failure :(

scottyschup:
> Manually QA-ed and this is looking good, but we have a test failure :(

Huh. They were all passing before I pushed that last little view-logic change which has nothing to do with that test failure. I'll try backmerging master.

Issue #349, opened 2020-02-09
Add user managment for MAC users - Part 1


jel535s:
## Context
- Need to add a new tier of MAC Admin users  
- Need to rename current tier of MAC users from "admin" to "coordinators"
- These new higher tier Admin users should have a different tier of access than coordinators.
- New tier users should be able to see MAC user information about login history / access logs
- New tier users should be able to add other MAC users through a UI

## Process
This ticket covers:
Building out the 2 tiers of MAC users as explained in Context above

## Acceptance Criteria/Definition of Done
Addition of a coordinator role, and movement of most existing MAC users to that role

Tests: 
Try to access this page as a coordinator level MAC login on the app (should **not** be able to access)
Try to access this page as an new Admin level MAC login on the app (**should** be able to access)

## Resources
"manage a partner" page for reference:
https://projects.invisionapp.com/share/9EMYOUG7ZBA#/screens/365422459

Mockups:
For the menu (with the "Admin Dashboard" button): https://invis.io/9EMYOUG7ZBA#/408360708_dashboard_-_MAC_Admin_-menu_Open-

For the admin dashboard: https://invis.io/9EMYOUG7ZBA#/408360709_dashboard_-_MAC_Admin_-admin_Dashboard-

jel535s:
@gujacq can you add a design here for the new MAC Admin dashboard view as outlined above?  Let me know if you need any further clarification, I know this is basically a rough outline.

gujacq:
view log in history:
- IP address
- Last logged in attempt (possibly drills down to last ten login attempts)
- Might not need a front end design for this

jel535s:
Updated ask for design based on IRL conversation:
- Build a view for where to access the new Admin Dashboard (possibly a nav button)
- Build a view of the new Admin Dashboard, including:
  - add a new MAC user and determine their role (admin vs. coordinator)
A list of all MAC users which includes:
  - lock/unlock MAC users
  - view last log in and last log in IP for every user
  - the status of the user's account (locked/pending/active)
  - their current role and the ability to change it (between admin and coordinator)

jel535s:
Spoke to Nicole and clarified that eventually this dashboard will also include finance reporting.

gujacq:
Designs here:

For the menu (with the "Admin Dashboard" button): https://invis.io/9EMYOUG7ZBA#/408360708_dashboard_-_MAC_Admin_-menu_Open-

[...]

PR #454, opened 2020-06-28
349 user management

misternu:
## Changelog
- Added an admin dashboard for managing admin users

## Link to issue:  
#349

## Relevant Screenshots: 
Screen Shot 2020-06-28 at 6 51 52 PM

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/app/javascript/src/admin_dashboard.js
+++ b/app/javascript/src/admin_dashboard.js
@@ -17,12 +17,24 @@ const handleUserResponse = response => {
   }
 }
 
+const handleExpandUser = event => {
+  button = $(event.target)
+  hidden = button.parents('tr').find('p:last-child')
+  hidden.slideToggle(300)
+  if (button.html() == "Expand record") {
+    button.html("Collapse record")
+  } else {
+    button.html("Expand record")
+  }
+}
+
 const setupAdminDashboard = () => {
-  const users = $('.admin-dashboard-users')
+  const users = $('.admin-dashboard-users-wrapper')
   if (users) {
-    console.log('setting up user form handler')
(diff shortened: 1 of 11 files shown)


app/javascript/src/admin_dashboard.js
```diff
@@ -0,0 +1,43 @@
...
+  form.reset()
+}
+
+const handleUserResponse = response => {
+  const data = response.detail[0]
+  if (data.users) {
+    refreshUsers(data.users)
+    clearUserForm()
+  } else {
+    console.log(data)
```
BintLopez, line 16: We should rm this so we don't leak data into the console in the absence of users

app/views/admin_dashboard/index.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    <%= f.label :name %>
+    <%= f.text_field :name, class: 'form-control' %>
+  </div>
+  <div class="form-group">
+    <%= f.label :email %>
+    <%= f.text_field :email, class: 'form-control' %>
+  </div>
+  <div class="form-group">
+    <%= f.label :role %>
+    <%= f.select :role, options_for_select(User::ROLES.reject { |r| r == 'partner' }), {}, class: 'form-control' %>
```
BintLopez, line 15: Instead of rejecting, could we do `[User::ADMIN]`? That way we keep selection logic out of the view until we support other types of user roles?

app/javascript/src/admin_dashboard.js
misternu: oops... 😅

PR #475, opened 2020-08-23
Balance tooltip 406

### LINKED ISSUE #406: Tool Tip for balance on Partner Details page
## Context
As a partner or MAC user, I may be confused about how the app is calculating the balance on the partner details page.

## Process
As a MAC user: sign in, then click on "View Partner Details".  Notice the balance in the upper right hand of the page.
As a partner user:  sign in, and notice on the front page that there is a balance in the upper right hand of the page.

Add a tool tip icon next to the balance.  This will be a clickable icon with text that pops up when clicked.  We prefer a clickable icon rather than a hover-over text, for mobile and accessibility reasons.

The copy when the icon is clicked should read: "The balance here includes all cash in the partner lockbox, minus any Pending Support Requests."

## Resources
Invisio designs for this are in comments below.

Screenshot of current prod instance of app, from MAC user: 
image
Screenshot of current prod instance of app, from a partner user:
image

kyaroch:
## Changelog
- Added tooltip to balance with explanatory text

## Link to issue:  
Fixes #406 

- N/A

## Relevant Screenshots: 
Screenshot_20200830_171526

Note: It's not possible to make the background of the tooltip arrow white, because it's not possible to put borders on it. The way Bootstrap implements this, the borders of the tooltip are the triangular areas outside the arrow, so coloring the borders would color those entire areas.

## Are you ready for review?:

- [ ] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/app/assets/stylesheets/bootstrap-overrides.scss
+++ b/app/assets/stylesheets/bootstrap-overrides.scss
@@ -77,3 +77,10 @@ a {
 .alert {
   margin-top: 1rem;
 }
+
+.tooltip > .tooltip-inner {
+  background-color: white;
+  color: black;
+  border: 1px solid $color-primary;
+  text-align: left;
+}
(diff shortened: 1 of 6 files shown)


BintLopez (approved):
🦑

PR #476, opened 2020-08-30
Change reconciliation guard clause from return to next

misternu:
should fix error described here https://github.com/MidwestAccessCoalition/lockbox_rails/issues/387#issuecomment-683477433


misternu:
@BintLopez removing performance requirement, the test at least checks to make sure those emails aren't sent synchronously which I guess is enough

Issue #368, opened 2020-02-17
System generated notes for updated support requests


BintLopez:
## Context
A few refinements of notes:
- Add punctuation at the end of each change noted in the update summary note text
- Show who made the update in the note text
- Handle updates to/from nil values or empty strings (play around with changing the urgency flag on a support request)

<img width="1163" alt="image" src="https://user-images.githubusercontent.com/5728859/74688353-e5c71900-519c-11ea-98ed-00e461e41a22.png">

## Process
Update a support request and observe the notes that are generated.

## Acceptance Criteria/Definition of Done
- notes text should have punctuation for legibility
- notes should indicate _who_ updated the support request
- display "blank" for updates to/from nil or empty string

gujacq:
@AG-Sile Can you provide an update to this ticket and any blockers you might have?

jel535s:
@AG-Sile was working on this ticket, but we're opening it back up to the group to grab. Not sure where they left off so it might be best to consider this a fresh start.

PR #488, opened 2020-09-20
Refining system generated notes for support requests

nsiwnf:
## Changelog
- Add punctuation at the end of each change noted in the update summary note text
- Include "Changes made by ___" in note text
- Displays "blank" for changes to and from nil/empty string values

## Link to issue:  
#368 

Sign in as admin and edit support requests to see generated notes

## Relevant Screenshots: 
Screen Shot 2020-09-20 at 3 06 17 PM

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/lib/update_support_request.rb
+++ b/lib/update_support_request.rb
@@ -60,10 +60,10 @@ class UpdateSupportRequest
 
     original_values.each do |field, original_value|
       new_value = support_request.send(field)
-      next unless new_value != original_value
-
       new_value = 'blank' if new_value.blank?
       original_value = 'blank' if original_value.blank?
+      next unless new_value != original_value
+
       note_text << "The #{field_labels[field]} for this Support Request was changed from "\
                     "#{original_value} to #{new_value}."
     end
--- a/lib/update_support_request.rb
+++ b/lib/update_support_request.rb
@@ -27,7 +27,7 @@ class UpdateSupportRequest
   # }
   #
   # We'll pass it directly to SupportRequest to take advantage of AcceptsNestedAttributes.
-  input :support_request, :params
+  input :support_request, :params, optional: [:current_user]
 
   attr_accessor :support_request, :original_values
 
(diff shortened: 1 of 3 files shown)


lib/update_support_request.rb
```diff
@@ -60,11 +60,15 @@ def notate_changes
 
     original_values.each do |field, original_value|
       new_value = support_request.send(field)
-      if new_value != original_value
-        note_text << "The #{field_labels[field]} for this Support Request was changed from #{original_value} to #{new_value}"
-      end
+      next unless new_value != original_value
+
+      new_value = 'blank' if new_value.blank?
+      original_value = 'blank' if original_value.blank?
```
nsiwnf, line 66: If someone changes something from `''` to `' '` - it will show as "changed blank to blank" - Should these 2 lines be moved above the `next` on line 63?

BintLopez: oh yes, good catch there. I think that's a good idea to move above ln 63

lib/update_support_request.rb
```diff
@@ -60,11 +60,15 @@ def notate_changes
...
-      end
+      new_value = 'blank' if new_value.blank?
+      original_value = 'blank' if original_value.blank?
+      next unless new_value != original_value
+
+      note_text << "The #{field_labels[field]} for this Support Request was changed from "\
+                    "#{original_value} to #{new_value}."
     end
 
+    note_text << "Changes made by #{support_request.user.name}." if support_request.user
```
BintLopez, line 71: Ooooof sorry I am just catching this... but the person who's associated with the support_request is not necessarily going to be the person making changes to a support request.

The user attached to a given support request is the coordinator that filed/created that support request. However, if a lockbox_partner user updates the support request status, this code will attribute that change to the coordinator attached to the support request rather than the user who actually made the updates. Rather than looking at the user who's on the support request, we should be figuring out which user prompted this update action.

spec/lib/update_support_request_spec.rb
```diff
@@ -73,10 +73,11 @@
       }
     }
 
-    expect{UpdateSupportRequest.call(support_request: support_request, params: update_params)}
+    expect{UpdateSupportRequest.call(support_request: support_request, params: update_params, current_user: support_request.user)}
```
BintLopez, line 76: Do we have a test condition where it's not passed in? Or where it's passed in but nil?

nsiwnf: Yup! In the tests on lines 140 and 172 - `current_user` is not passed

BintLopez (approved):
🦑

Issue #464, opened 2020-08-02
Add acceptance test covering pagination behavior


BintLopez:
## Context
This is a follow up to https://github.com/MidwestAccessCoalition/lockbox_rails/pull/407, which added pagination to the app for historical lockbox actions. While we don't need to test the ins and outs of pagination, as that should be covered by the gem, we should add an acceptance test to ensure that pages load properly when there are enough lockbox actions to paginate.

## Process
Add an acceptance test to ensure that page loads properly when there are enough lockbox actions to paginate.

## Acceptance Criteria/Definition of Done
Simple test of a paginated page exists.

PR #504, opened 2020-10-24
464 Add acceptance tests for lockbox partner show page pagination

misternu:
## Changelog
- Add acceptance tests for lockbox partner show page pagination

## Link to issue:  
#464

### DIFF
--- a/spec/system/lockbox_partner_pagination_spec.rb
+++ b/spec/system/lockbox_partner_pagination_spec.rb
@@ -1,0 +1,22 @@
+require 'rails_helper'
+
+RSpec.describe "LockboxPartner show page", type: :system do
+  let!(:lockbox_partner) { create(:lockbox_partner, :active) }
+  let!(:user) { create(:user) }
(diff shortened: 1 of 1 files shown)


spec/system/lockbox_partner_pagination_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+
+RSpec.describe "LockboxPartner show page", type: :system do
+  let!(:lockbox_partner) { create(:lockbox_partner, :active) }
+  let!(:user) { create(:user) }
+  before do
+    # 1 completed add cash action already exists
+    19.times { create(:support_request, :pending, lockbox_partner: lockbox_partner) }
```
BintLopez, line 8: All this data creation is going to make for a slow test.... but... I say we refactor and deal with slow test suite when it's a problem and err on the side of more robust testing :)

BintLopez (approved):
🦑

PR #505, opened 2020-10-25
Fix support request controller spec time zone flake

misternu:
To reproduce this spec error (https://app.circleci.com/pipelines/github/MidwestAccessCoalition/lockbox_rails/1137/workflows/2715855b-e551-448a-acb8-396a3f7ca7bc/jobs/1326), run support_requests_controller_spec in current trunk with the following inserted at line 31 `Timecop.travel(DateTime.new(2020, 10, 20, 0, 30, 0))`

`SupportRequest.to_csv` is computed in UTC when called on line 26, but when we call #export in the spec it's being computed in the Chicago time zone (user factory always sets Chicago time zone). This will make the spec fail for 5 hours every day from 7 pm to midnight Chicago time.

The fix is on line 26 of this PR, ensuring that the test string is being computed in the user's time zone. The rest is just cleaning up calls to user factories.

### DIFF
--- a/spec/controllers/support_requests_controller_spec.rb
+++ b/spec/controllers/support_requests_controller_spec.rb
@@ -4,8 +4,8 @@ describe SupportRequestsController do
 
   describe "#index" do
     context "when the user is an admin" do
+      let(:user) { create(:user) }
       it "returns 200" do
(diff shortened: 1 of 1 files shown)


BintLopez (approved):
🦑

PR #511, opened 2020-11-22
Expense categories 490

### LINKED ISSUE #490: Match Support Request expense categories with MAC Finance Quickbooks categories
## Context
Each expense category for Support Requests should correspond with an accounting code in Quickbooks. This will help our Finance team better track and report on client expenses.

## Process
Currently on the Support Request form, there is a drop-down list of expense categories which map to Lockbox Transactions.  This list is _close_ to our MAC Finance team's actual Quickbooks financial categories used for reporting, but with some key differences. Differences are listed in Acceptance Criteria.

Ideally, we would also want to build some code in that when an expense category is selected, the corresponding Quickbooks category code is tracked as well for financial reporting as built in #351 . The category codes are listed in the excel doc attached in Resources below.

On the Support Request form, the list of categories which can be line items/Lockbox Transactions should match exactly the list of relevant Quickbooks categories. Below are the differences. Attached in Resources below full list of Quickbooks categories, compared to the current Lockbox options.

- [ ] In the Lockbox, we have "Transit".  This should be broken out to be: Bus Travel, Train Travel, Rideshare Travel, and Air Travel. 

- [ ] In the Lockbox, we have "Gas". This should be changed to "Gas Reimbursement".

- [ ] In the Lockbox, we have "Parking". This should be changed to "Other Client Travel- Parking".

- [ ] In the Lockbox, we have " Hotel Reimbursement". This should be changed to "Accommodations".

- [ ] In the Lockbox, we have "Medicine".  This should be changed to "Prescriptions".

- [ ] In the Lockbox, we have "Childcare".  This should be changed to "Childcare Reimbursements".

- [ ] In the Lockbox, we have two categories that do not match a current category code in Quickbooks: "Adjustment" and "Cash Addition".  These should be left as-is currently, since they track Lockbox-specific behavior.  They may end up being re-adjusted after the first year of Finance reporting.

- [ ] We should add options for "Other Client Supplies" and "Procedures" in the Lockbox Support Requests.  Although these aren't Lockbox-specific needs yet, Finance indicated they would like these tracked and that matches with our guidance from leadership about the direction the app is headed.

- [ ] Note- "Food" is unchanged, since the category name is the same in both Quickbooks and Support Requests in the app.

## Resources
[Lockbox App and MAC.Expense.Account.Categories.xlsx](https://github.com/MidwestAccessCoalition/lockbox_rails/files/5252834/Lockbox.App.and.MAC.Expense.Account.Categories.xlsx)

kyaroch:
## Changelog
- Add new QuickBooks expense categories
- Redefine existing categories (except cash addition, adjustment, and food) as legacy categories, and don't allow them to be selected through the UI

## Link to issue:  
Fixes #490 

- See comments

## Relevant Screenshots: 
Can't screenshot the dropdown itself on this computer, but here are representative selections:

Screenshot_20201122_170149

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue


app/models/lockbox_transaction.rb
```diff
@@ -17,23 +17,71 @@ class LockboxTransaction < ApplicationRecord
...
+    PROCEDURES                  = 'procedures',
+  ].freeze
+
+  INTERNAL_EXPENSE_CATEGORIES = [
+    # Used for internal Lockbox purposes
+    ADJUSTMENT    = 'adjustment',
+    CASH_ADDITION = 'cash_addition',
+  ].freeze
+
+  LEGACY_EXPENSE_CATEGORIES = [
```
kyaroch, line 42: I'm assuming we _don't_ want to make existing records with these categories invalid. If we do, we should probably do a data migration.

BintLopez: Good looking out here. I think this PR is good to go to stop creation of less than idea data, but I think we're going to need to do a migration to clean up the old data. This will be especially important for any future reporting functionality. cc/ @gujacq @jel535s

app/models/lockbox_transaction.rb
```diff
@@ -17,23 +17,71 @@ class LockboxTransaction < ApplicationRecord
...
+    PROCEDURES                  => '6012',
+  }.freeze
+
   def eff_date
     lockbox_action.eff_date
   end
 
+  def quickbooks_category_code
+    QUICKBOOKS_CATEGORY_CODES[category]
+  end
```
scottyschup, line 83: I didn't see this used anywhere. Is it just anticipating some future need?

scottyschup (commented):
Just had one question.

app/models/lockbox_transaction.rb
kyaroch: Yes, this is meant to allow the category code to be populated when generating reports (as specified in the issue)

app/models/lockbox_transaction.rb
kyaroch: I'd just add the migration now, but I'm not sure what we're going to do about the "Travel" category. We have no way of knowing which of the travel categories it should be migrated to.

BintLopez:
We need to update the seeds file to create only valid expenses so that our staging app data and development data are working off the same assumptions as the rest of the app
- https://github.com/MidwestAccessCoalition/lockbox_rails/blob/trunk/db/seeds.rb#L63
- https://github.com/MidwestAccessCoalition/lockbox_rails/blob/trunk/db/seeds.rb#L94

BintLopez (approved):
🦑

app/models/lockbox_transaction.rb
BintLopez: We can handle the migration piece separately after checking about what category we should use for the old travel category.

PR #521, opened 2020-11-29
Redact pii in app 467

### LINKED ISSUE #467: Redact PII in App
## Context
As part of the Lockbox MVP, we need to automatically redact PII from the app and databases after 3 months.

## Process
Using the research from #468, it was decided that we are not logging PII anywhere other than Client Name/Alias ID fields, and possibly user notes.

When a completed/canceled status case is older than 3 months old, the following should be redacted automatically:
- Client name
- Client Alias
- All user-generated notes
- Name changes _could_ also be captured in system generated notes if the user changes the client name or alias fields after generating a Support Request- and that particular type of system-generated note also needs to always be redacted. 

You are 100% allowed to build in fun easter-egg style replacements for PII, instead of just using "redacted".

This would likely be a nightly cron job running a redaction script. This must run on off-hours as it will likely be a gnarly query.

kyaroch:
## Changelog
- Added methods for redacting support requests and notes
- Added `redacted` column to indicate whether a support request has been redacted
- Created scheduled job to redact all support requests that were closed over 90 days ago

## Link to issue:  
Fixes #467 

- Verified that `redact_async` does enqueue a Sidekiq job, although the queues don't appear to run on my local machine

## Relevant Screenshots: 
Screen Shot 2020-12-13 at 5 15 42 PM

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue


app/models/note.rb
```diff
@@ -23,6 +27,10 @@ def system_generated?
     user_id.blank?
   end
 
+  def redact!
+    update!(text: SupportRequest::REDACTED)
```
kyaroch, line 31: At this point I don't think it's worthwhile to track the redaction status of individual notes, since they only get redacted when the SupportRequest is redacted. In the worst case scenario, they'll get redacted twice.

nsiwnf (approved):
This lgtm - I guess the schema changes could go in a separate deploy, but it's not necessary 🦑

app/models/support_request.rb
```diff
@@ -28,6 +28,18 @@ class SupportRequest < ApplicationRecord
     scope "#{status}_for_partner", ->(lockbox_partner_id:) { joins(:lockbox_action).where(lockbox_partner_id: lockbox_partner_id, "lockbox_actions.status": status) }
   end
 
+  REDACTED = "[redacted]".freeze
+  REDACT_AFTER_DAYS = 90
+
+  scope :needs_redaction, -> do
```
BintLopez, line 34: We should add some tests for this query

app/models/lockbox_action.rb
```diff
@@ -20,16 +20,18 @@ class LockboxAction < ApplicationRecord
   before_validation :set_default_status
   has_paper_trail
 
+  before_save :record_closed_at, if: :will_save_change_to_status?
```
BintLopez, line 23: Today I learned about `:will_save_change_to_` .... very cool!

app/models/note.rb
```diff
@@ -7,6 +7,10 @@ class Note < ApplicationRecord
 
   validates :notable_action, inclusion: { in: %w{create update annotate} }
 
+  # User-generated notes could contain anything, so they're included in this
+  # scope regardless of whether may_contain_pii is set
+  scope :may_contain_pii, -> { where("user_id IS NOT NULL OR may_contain_pii") }
```
BintLopez, line 12: @jel535s @gujacq -- so for the redaction task, we're treating all user generated notes as if they have PII in them, which is the safest default behavior since we have no control over what's in those notes. However, since we're adding a column in the database to track this, this would enable us on the backend to, say, allow a note author to mark their note as not containing pii so as to avoid redaction. Wondering if we'd want to build on this backend functionality to allow a checkbox at the note field to not redact a note. This is not a blocking comment for this PR but wanted to point out that we could support that functionality should we see fit. I'd imagine that being useful if we want to enter certain notes into the permanent record for whatever reason.

BintLopez: Welp. I read this incorrectly... 🤡

BintLopez: Okay, so if we were to backpopulate all user-generated notes with may_have_pii: true, then we could get rid of the user_id = null part of the query and just go based off of whether it's been marked as potentially PII, so if we did want to skip redaction in the future it would just take altering this query and adding a field to the notes UI

lib/update_support_request.rb
```diff
@@ -58,18 +58,25 @@ def cache_original_values
...
 
+    may_contain_pii = false
+
     original_values.each do |field, original_value|
       new_value = support_request.send(field)
       new_value = 'blank' if new_value.blank?
       original_value = 'blank' if original_value.blank?
       next unless new_value != original_value
 
+      may_contain_pii = true if field == :name_or_alias
```
BintLopez, line 69: This should have some test coverage. Let's ensure update support request note creation sets this field properly with a test

BintLopez (approved):
🦑 thanks for adding test coverage! This LGTM!

Though there is something that occurred to me (that we can address in a followup PR)... we are using the papertrail gem to make an audit trail / changelog of database changes of the core tables. We're using that on SupportRequest and thus the PII is also going to be logged to versions. We could either stop logging to versions for that table OR we could look into preventing logging of certain fieldds in the versions table.

Issue #483, opened 2020-09-06
Mobile - Partner user- Alerts


jel535s:
## Context
As our users are often accessing the app on mobile devices, we did an audit for usability on #424

## Process
When signed in as a Partner user, you might see the following alerts: Reconciliation Due, Cash Sent/confirm cash, or the low balance alert. 
The low balance alert seems to have a re-sizing issue on mobile, where the text is overflowing outside of the background alert color box.  This isn't a functionality blocker, but we'd like to get it to match the other alerts and stay inside the box.
Screenshots in Resources are from an iPhone SE (4.7 inch screen) in the Firefox browser.

## Acceptance Criteria/Definition of Done
Text for the low balance alert remains inside the box.

## Resources
Current mobile view:
<img width="375" alt="Partner alerts" src="https://user-images.githubusercontent.com/53353956/92336580-ce337d00-f067-11ea-9979-423cbfef167c.png">
<img width="372" alt="Partner user alert messages text too wide" src="https://user-images.githubusercontent.com/53353956/92336582-cffd4080-f067-11ea-91af-a6f6128e98da.png">

PR #524, opened 2020-12-06
Adding break-word for text overflow in alerts

nsiwnf:
## Changelog
- Added css to "break words" if the text overflows

## Link to issue:  
Fixes issue #483 

- N/A

## Relevant Screenshots: 

Before:
before
After: 
after

## Are you ready for review?:

- [ ] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/app/assets/stylesheets/alert.scss
+++ b/app/assets/stylesheets/alert.scss
@@ -6,7 +6,3 @@
   max-width: 600px;
   padding: 0;
 }
-
-.break-word {
-  word-break: break-word;
-}
--- a/app/assets/stylesheets/alert.scss
+++ b/app/assets/stylesheets/alert.scss
@@ -6,3 +6,7 @@
   max-width: 600px;
   padding: 0;
 }
+
+.break-word {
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/alert.scss
```diff
@@ -6,3 +6,7 @@
   max-width: 600px;
   padding: 0;
 }
+
+.break-word {
```
BintLopez, line 10: Instead of adding our own class, it might be better to use Bootstrap's existing class for this -- https://getbootstrap.com/docs/4.3/utilities/text/#word-break

It looks like it might also provide a bit more coverage for other browsers (like IE)

BintLopez: It also sets `overflow-wrap` in addition to `word-break` so let's use the Bootstrap class if it works instead of this one.

nsiwnf: Ah, good call - will make that change!

misternu (approved):
🦉

PR #551, opened 2021-01-18
Refactor test to fix intermittent failure

### LINKED ISSUE #530: Investigate and improve flaky test - pagination
## Context
The `spec/system/lockbox_partner_pagination_spec.rb` test fails intermittently, only finding 19 rather than 20 items. This is likely due to the way that the test is creating data in its set up. Perhaps a different strategy would be to create less test data but stub out the pagination setting that caps the items on a page to 20 such that pagination would occur for more than 1 or 2 items on a page.

## Process
- Reproduce the failure
- Fix the test such that it no longer fails intermittently

- Intermittent issue is reproduced and understood
- Test no longer fails intermittently

## Resources

kyaroch:
## Changelog
- Refactored test into two tests – I thought this might fix the issue, and it didn't, but since we're testing two scenarios here, it's better to test them in separate tests.
- Made test less specific. This intermittent failure appears to be caused (at least in part) by the initial add_cash lockbox action not being completed in some cases, when the entire test suite is run. It's unclear to me why this happens. By specifying _exactly_ what should be displayed on both pages, we're effectively testing the factories and the state of the database, which is a lot more than this spec is intended to test; it should only test pagination itself.

## Link to issue:  
Fixes #530 

- N/A

## Relevant Screenshots: 
- N/A

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/spec/system/lockbox_partner_pagination_spec.rb
+++ b/spec/system/lockbox_partner_pagination_spec.rb
@@ -3,21 +3,27 @@ require 'rails_helper'
 RSpec.describe "LockboxPartner show page", type: :system do
   let!(:lockbox_partner) { create(:lockbox_partner, :active) }
   let!(:user) { create(:user) }
-  before do
-    # 1 completed add cash action already exists
-    19.times { create(:support_request, :pending, lockbox_partner: lockbox_partner) }
-    login_as(user, :scope => :user)
-  end
 
-  it "has pagination" do
-    visit("/lockbox_partners/#{lockbox_partner.id}")
-    page.assert_selector(".support-request-container ul.pagination", count: 0)
-    create(:support_request, :pending, lockbox_partner: lockbox_partner)
-    visit("/lockbox_partners/#{lockbox_partner.id}")
-    page.assert_selector(".support-request-container ul.pagination", count: 1)
-    # TODO: investigate why this test is so flaky (often only finds 19)
-    page.assert_selector(".support-request-container .lockbox-activity tr.pending", count: 20)
-    click_link("2")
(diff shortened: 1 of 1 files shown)


BintLopez (approved):
🦑

spec/system/lockbox_partner_pagination_spec.rb
```diff
@@ -3,21 +3,27 @@
...
+  context "when there are not enough support requests to require pagination" do
+    it "does not have any pagination" do
+      visit("/lockbox_partners/#{lockbox_partner.id}")
+      page.assert_selector(".support-request-container ul.pagination", count: 0)
+    end
   end
+ 
+  context "when there are enough support requests to require pagination" do
+    before do
+      20.times { create(:support_request, :pending, lockbox_partner: lockbox_partner) }
```
BintLopez, line 18: Creating a record 20 times is an expensive thing to do... but given the fact that our app and test suite are small and run quickly, I don't think this is an issue. More complete and realistic coverage is more important than test suite speed right now :) ... As our codebase grows though, we will likely want to revisit this test to make sure we don't unnecessarily bog down the test suite with a lot of data creation.

kyaroch: Agreed! There should be faster ways to do this when we need to.

misternu: I'm responsible for the original spec and I'm sorry for this. Perhaps we could stub support requests so that it returns fake data when rendering the page/pagination?

kyaroch: That seems like the way to go, but not our highest priority right now, IMO.

Issue #371, opened 2020-02-18
Make low balance trigger amount configurable per partner


jel535s:
## Context
Lockbox finance team needs low balance alerts to be set at different amounts for different partners, so that they can send out payments proactively.
Edited 12/6 to include the below comments in this top comment so that it is ready to work!

## Process

- [x] Confirm with lockbox finance team the current 3 partners' trigger balances.

When a partner's balance reaches a certain amount, an email and in-app notification should be sent to the lockbox and finance managers.  This trigger amount will need to vary, based on the partner.  
This trigger amount also needs to factor in pending requests as if they were completed (which is the current workflow for visible balance at the top of the partner details page).
Notes from 11/15 meeting: FPA balance should be set at $750. $500 will be the default for new partners. All partner lockboxes should have the ability to be changed by MAC leadership.
We decided not to implement the ability for partners to adjust their own low balance alerts.
We decided to make this part of MVP Lockbox.

- [ ] Need to build in a field that MAC users can edit on the "partner management" view and on the "add a new lockbox partner" view, where they can set this trigger amount.

- [x] Need to design where this field shows up for MAC users -- in the onboarding form and in the partner management panel

- [x] Need to design how this shows up for a partner view so they're informed what the balance is (but cannot change it)

## Acceptance Criteria/Definition of Done

- [ ]  In the Manage Lockbox partner view, add a toggle where MAC users will set the "low balance trigger amount" for each partner on an ongoing basis.  

- [ ] On the Add a new lockbox partner" page at https://demo-lockbox.herokuapp.com/lockbox_partners/new add a required field to set the low balance trigger amount, as partners are created.  Field title copy should be "Lockbox Low Balance Alert Amount" with instruction text below that reads "When the lockbox balance reaches this amount, notifications will be sent to the lockbox finance team and coordinators to send out more money."

## Resources
This is related to https://github.com/MidwestAccessCoalition/lockbox_rails/issues/451

New design for Manage Lockbox Partner view:
https://invis.io/9EMYOUG7ZBA#/365422459_dashboard_-_MAC_User_-invite_Lockbox_Partner_User-

New design for Add New Lockbox Partner view:
https://invis.io/9EMYOUG7ZBA#/365424955_dashboard_-_MAC_-add_New_Lockbox_Partner-

jel535s:
From Kate/Veronica, Low balance triggers should be set at:
FPA: $750
All other: $500

gujacq:
[...]

PR #556, opened 2021-01-24
1/2 Add new minimum_acceptable_balance column with default value of 50000

nsiwnf:
## Changelog
Adding a new column `minimum_acceptable_balance` with default value of 50000. 

This is the first part of #371 - if there are any issues migrating this column after it's being referenced, then it might leave the application in an unstable state. 

## Link to issue:  
#371 

- This column is not used in this PR. It is just the migration

## Relevant Screenshots: 
- N/A

## Are you ready for review?:

- [ ] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/db/migrate/20210117223939_add_minimum_acceptable_balance_to_lockbox_partners.rb
+++ b/db/migrate/20210117223939_add_minimum_acceptable_balance_to_lockbox_partners.rb
@@ -1,0 +1,5 @@
+class AddMinimumAcceptableBalanceToLockboxPartners < ActiveRecord::Migration[6.0]
+  def change
+    add_column :lockbox_partners, :minimum_acceptable_balance, :integer, default: 50000
+  end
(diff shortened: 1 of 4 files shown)


db/migrate/20210117223939_add_minimum_acceptable_balance_to_lockbox_partners.rb
```diff
@@ -0,0 +1,5 @@
+class AddMinimumAcceptableBalanceToLockboxPartners < ActiveRecord::Migration[6.0]
+  def change
+    add_column :lockbox_partners, :minimum_acceptable_balance, :integer, default: 50000
```
BintLopez, line 3: let's move this default amount from the migration to the model

BintLopez (approved):
🦑 lgtm!

PR #557, opened 2021-01-24
Prevent paper_trail from recording name_or_alias

### LINKED ISSUE #555: Discovery and implementing: Remove PII in papertrail gem
## Context
We are using the papertrail gem to make an audit trail / changelog of database changes of the core tables. We're using that on SupportRequest and thus the PII is also going to be logged to versions. We could either stop logging to versions for that table OR we could look into preventing logging of certain fields in the versions table.
Related to #467 

## Process
Investigate the pros and cons of ending logging to versions for SupportRequest vs preventing logging of certain fields in the versions table. Choose the best way to prevent PII from being logged to versions, and implement that choice.

PII is no longer logged in versions via the papertrail gem.

kyaroch:
## Changelog
- Configure paper_trail to ignore the `support_requests.name_or_alias` column

## Link to issue:  
Fixes #555 

I verified in the console that paper_trail doesn't record `name_or_alias` when a SupportRequest is created or updated. If an update changes only the `name_or_alias`, that update isn't recorded in the versions table at all, which I assume is fine.

We do not use paper_trail for notes, so this is the only field we need to be concerned about.

## Relevant Screenshots: 
- N/A

## Are you ready for review?:

- [ ] Added relevant tests
- [ ] Linked PR to the issue

### DIFF
--- a/app/models/support_request.rb
+++ b/app/models/support_request.rb
@@ -17,7 +17,8 @@ class SupportRequest < ApplicationRecord
 
   # Sometimes the UUID will already have been created elsewhere, and sometimes not
   before_validation :populate_client_ref_id
-  has_paper_trail
+  # Do not save possible PII to the versions table
+  has_paper_trail ignore: [:name_or_alias]
 
   # for grepability:
   # scope :pending
--- a/spec/models/support_request_spec.rb
+++ b/spec/models/support_request_spec.rb
@@ -268,4 +268,34 @@ describe SupportRequest, type: :model do
       expect(note3.reload.text).not_to eq(SupportRequest::REDACTED)
     end
   end
+
+  describe "paper_trail" do
+    let(:support_request) do
+      create(:support_request, name_or_alias: "old name", urgency_flag: "old flag")
+    end
+
+    context "on creation" do
+      it "does not record name_or_alias" do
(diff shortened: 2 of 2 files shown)


app/models/support_request.rb
```diff
@@ -17,7 +17,8 @@ class SupportRequest < ApplicationRecord
 
   # Sometimes the UUID will already have been created elsewhere, and sometimes not
   before_validation :populate_client_ref_id
-  has_paper_trail
+  # Do not save possible PII to the versions table
+  has_paper_trail ignore: [:name_or_alias]
```
BintLopez, line 21: @kyaroch -- can we add a test for this in the support request model? Basically to ensure that name_or_alias does not exist in the versions changelog? (if we're not running papertrail in test... no need)

spec/models/support_request_spec.rb
```diff
@@ -268,4 +268,34 @@
       expect(note3.reload.text).not_to eq(SupportRequest::REDACTED)
     end
   end
+
+  describe "paper_trail" do
```
BintLopez, line 272: THANK YOU!

BintLopez (approved):
🦑

Issue #369, opened 2020-02-18
Lockbox Partners appearing twice in file support request form dropdown


BintLopez:
## Context
Lockbox partners appear twice in the file support request dropdown
<img width="1032" alt="image" src="https://user-images.githubusercontent.com/5728859/74763884-38a8db00-5246-11ea-878e-bc0594c32185.png">

## Process
- From the admin user dashboard, go to file a new support request (support_requests/new)
- Observe the duplicate partner listings.

## Acceptance Criteria/Definition of Done
- Partners should not be listed twice in the dropdown

BintLopez:
It's back! The bug has reared its head again! 🐜 🐝 🐞 🐛

PR #563, opened 2021-01-25
Add extra user per lockbox partner to seeds for local testing

misternu:
This will ensure that the support request dropdown issue will be locally testable, should it return. low priority

Follow up from https://github.com/MidwestAccessCoalition/lockbox_rails/issues/369


BintLopez (approved):
🦑

PR #564, opened 2021-01-25
Redact support requests closed exactly 90 days ago

kyaroch:
## Changelog
- Support requests will be included in the `needs_redaction` scope if they were created _exactly_ 90 days ago. This is probably the correct behavior (although it's unlikely to come up in prod) and fixes some intermittent test failures.

## Link to issue:  
Fixes issue #N/A

- N/A

## Relevant Screenshots: 

## Are you ready for review?:

- [ ] Added relevant tests
- [ ] Linked PR to the issue

### DIFF
--- a/app/models/support_request.rb
+++ b/app/models/support_request.rb
@@ -38,7 +38,7 @@ class SupportRequest < ApplicationRecord
         support_requests: { redacted: false },
         lockbox_actions: { status: LockboxAction::CLOSED_STATUSES }
       )
-      .where("lockbox_actions.closed_at < ?", REDACT_AFTER_DAYS.days.ago)
+      .where("lockbox_actions.closed_at <= ?", REDACT_AFTER_DAYS.days.ago)
   end
 
   def all_support_requests_for_partner


app/models/support_request.rb
```diff
@@ -38,7 +38,7 @@ class SupportRequest < ApplicationRecord
         support_requests: { redacted: false },
         lockbox_actions: { status: LockboxAction::CLOSED_STATUSES }
       )
-      .where("lockbox_actions.closed_at < ?", REDACT_AFTER_DAYS.days.ago)
+      .where("lockbox_actions.closed_at <= ?", REDACT_AFTER_DAYS.days.ago)
```
BintLopez, line 41: Can we change the test instead of changing the after 90 days behavior?

BintLopez: Actuallllllyyyy nvm. Including records on the 90 day mark is actually more in line with how we're redacting PII as an org :D

BintLopez (approved):
🦑

Issue #520, opened 2020-11-29
Round up Support Requests to the next dollar


jel535s:
## Context
Behavior note: clinics always round up to the next dollar amount if a Support Request is for an amount that includes change.

## Process
We need to build in a calculated amount which rounds up (and only up, **never** down) to the next dollar for Support Requests. This will make our balance tracking more accurate, since it will more closely track what clinics are currently doing in real life.

This rounding should also be noted by a System Note. 

System Note text: _This Support Request was rounded up to the nearest dollar to accommodate cash disbursement through Lockbox._ 

## Acceptance Criteria/Definition of Done
Any Support Request that generates a decimal amount (for example, due to a gas mileage reimbursement) is automatically rounded up to the next dollar. 

## Resources
n/a

PR #574, opened 2021-03-06
520 Round up all support requests to the nearest dollar

misternu:
## Changelog
- Round the amount of a lockbox action up to the nearest dollar
- Because of this, reconciliation has to treat a lockbox action as a balance effect in and of itself
- Specs updated to match new behavior
- Made lockbox action rounding depend on `#round_up?` so that it can be configurable per tenant in the future
- Updated form behavior to match and note to the user that the rounding is happening
- Fixed buggy support request form behavior by updating dependencies and fixing gas reinbursement slug

## Link to issue:  
#520 

## Relevant Screenshots: 
Screen Shot 2021-03-06 at 4 07 10 PM

## Are you ready for review?:

- [x] Added relevant tests
- [x] Linked PR to the issue


app/javascript/packs/application.js
```diff
@@ -8,7 +8,7 @@ require('turbolinks').start();
 //require('@rails/activestorage').start();
 require('babel-polyfill');
 require('channels');
-require('cocoon-js');
+require("@nathanvda/cocoon")
```
BintLopez, line 11: is this a different version of that same dependency?

app/models/lockbox_action.rb
```diff
@@ -181,4 +183,9 @@ def record_closed_at
       self.closed_at = nil
     end
   end
+
+  def round_up?
+    # We may need to make this configurable when we go multi-tenant based on lockbox-partner
```
BintLopez, line 188: 🙏 bless -- good future proofing here

app/javascript/packs/application.js
misternu: looks like we were using a version of the cocoon js that someone other than nathan had made, from before when they added webpack instructions in the readme. Since then the gem must've got updated, but we had js that was mismatched with the changes to the gem. This is the official node package that goes with their ruby gem

app/models/lockbox_action.rb
```diff
@@ -103,7 +103,9 @@ def eff_date_formatted
   def amount
     return Money.zero if canceled?
     return Money.zero if lockbox_transactions.none?
-    lockbox_transactions.map(&:amount).sum
+    sum = lockbox_transactions.map(&:amount).sum
+    return sum unless round_up?
+    Money.new(sum.to_f.ceil * 100)
```
BintLopez, line 108: do we need to round up here if we are rounding when we're capturing the input? Thinking that as long as we're rounding up all the transactions that the amount shouldn't be rounding anything.

app/models/lockbox_partner.rb
```diff
@@ -33,14 +33,14 @@ def pending_support_requests
...
-        balance += action.amount
-      when LockboxTransaction::DEBIT
-        balance -= action.amount
+    relevant_actions_for_balance(exclude_pending: exclude_pending).inject(Money.zero) do |balance, action|
+      if action.credit?
+        balance + action.amount
+      elsif action.debit?
+        balance - action.amount
+      else
+        balance
```
BintLopez, line 42: Why was this change necessary?

app/views/support_requests/_form.html.erb
```diff
@@ -41,7 +41,10 @@
       </div>
     </div>
   <% end %>
-  <legend id="total">Total: $0.00</legend>
+  <div class="form-group">
+    <legend id="total">Total: $0.00</legend>
+    <p class="form-text text-muted font-italic my-0">Rounded up to the nearest dollar</p>
```
BintLopez, line 46: Should we put this <p> inside an if statement and only show when round_up? is true?

app/models/lockbox_action.rb
misternu: We aren't rounding up each transaction, only the whole support request

misternu:
Closed in favor of total rework, the parts of this PR which fix current prod bugs are cherry picked to #591

PR #581, opened 2021-03-20
fun gif

YongYi-coding:
Image is not rendering on the DOM. Please review and provide suggestions.  Thanks!!!


YongYi-coding:
@BintLopez Sorry I forgot to tag you last night.

config/webpack/environment.js
```diff
@@ -1,13 +1,3 @@
-const { environment } = require('@rails/webpacker');
```
BintLopez, line 1: I don't think this file needs to change. We want to make sure that we're still adding jQuery and Popper to webpack config

BintLopez (commented):
I'm noticing a few `.DS_Store` files checked in. We don't want to include these in the repo. I think it'd be great at this point to add `.DS_Store` to the `.gitignore` file so that we don't check these into the repo.

misternu:
it's possible that the javascript in thankyou_gifs.js is running immediately on pageload, where it might need to wait for document ready. Document ready in lockbox will look something like the following, you'll see this pattern in several other js files:

```
document.addEventListener('turbolinks:load', () => {
  // the page is ready to go, grab elements and do stuff to them
});
```

misternu:
the gif source also needed to be added to the content security policy, working example in this branch: https://github.com/MidwestAccessCoalition/lockbox_rails/compare/572-test

BintLopez:
Closing this PR in favor of https://github.com/MidwestAccessCoalition/lockbox_rails/pull/582

PR #582, opened 2021-03-22
fun gifs on login page

YongYi-coding:
@BintLopez @misternu Hello With Manu's help, the fun gifs logic is working now. I think the question is that how/where do we want to store those fun gifs? (I call them thank you gifs though). 

Right now I store the gif sources in an array. Manu pointed out that the gif source needs to be added to the content security policy like so `policy.img_src         :self, 'https://*.amazonaws.com', 'https://i.pinimg.com/'`. Since gif sources come from different urls, which means new developers who add their fun gifs will need to do the content security policy step. I should then write out an instruction for new dev to add fun gifs? 

If there's a solution that doesn't require to do the content security step to make the fun gif load, let me know.

Thanks!

### DIFF
--- a/app/views/layouts/devise.html.erb
+++ b/app/views/layouts/devise.html.erb
@@ -22,6 +22,7 @@
         </div>
         <%= yield %>
         <a id="sqreen-badge" title="Realtime application protection" href="https://www.sqreen.com/?utm_source=badge"><img style="width:109px;height:36px" src="https://s3-eu-west-1.amazonaws.com/sqreen-assets/badges/20171107/sqreen-light-badge.svg" alt="Sqreen | Runtime Application Protection" /></a>
+        <img src='' id='thankyou_gif' width='370px' alt='thank you gif' />
       </div>
     </div>
   </body>
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -366,7 +366,7 @@ GEM
       actionpack (>= 4.0)
       activesupport (>= 4.0)
       sprockets (>= 3.0.0)
-    sq_mini_racer (0.3.1.0.2)
+    sq_mini_racer (0.3.1.0.4)
     sqreen (1.22.1)
       libsqreen (~> 1.0)
       sq_mini_racer (~> 0.2, < 0.4.a)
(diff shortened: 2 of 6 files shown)


YongYi-coding:
Screen Shot 2021-03-26 at 4 26 15 PM
Here's how the log in page looks like with the fun gif. Let me know what should be the best styling (the the fun gif at the bottom or on the side)

BintLopez:
Thanks for flagging this question of how/where to store the gifs. I think our options are as follows...
- add gif files to the repo
- add each gif to the CSP allow list and fetch through the link
- only fetch gifs from giphy and add giphy to CSP allow list

app/views/layouts/devise.html.erb
```diff
@@ -22,6 +22,7 @@
         </div>
         <%= yield %>
         <a id="sqreen-badge" title="Realtime application protection" href="https://www.sqreen.com/?utm_source=badge"><img style="width:109px;height:36px" src="https://s3-eu-west-1.amazonaws.com/sqreen-assets/badges/20171107/sqreen-light-badge.svg" alt="Sqreen | Runtime Application Protection" /></a>
+        <img src='' id='thankyou_gif' width='370px' alt='thank you gif' />
```
BintLopez, line 25: The JS thank you gif functionality could also be done with ruby by using a helper method! For example you could write a method like the following in `app/helpers/application_helper.rb`....
```ruby
def random_thank_you_gif
  [
    'array',
    'of',
    'gifs'
  ].shuffle.last
end
```

And then in the html here you'd just need to call that helper method...

```html
<img src="<%= random_thank_you_gif % id='thankyou_gif' width='370px' alt='thank you gif'>"
```

misternu: https://ruby-doc.org/core-1.9.3/Array.html#method-i-sample

misternu:
when testing your original PR on mobile, having the gif underneath the way it was already looked great

YongYi-coding:
@BintLopez @misternu Here's the gif on the right side
Screen Shot 2021-03-28 at 3 36 44 PM

app/views/layouts/devise.html.erb
BintLopez: lol @misternu I knew I was forgetting an easy ruby method! Yessss #sample for the win

jel535s:
I agree that horizontal looks best on a desktop view! I'm concerned that horizontal on mobile will make the screen scroll. So that makes me lean towards the vertical option where the gif is below or above the login fields.

ElizabethHughes:
Looks like this was closed in favor of #590

PR #583, opened 2021-03-27
Environment issues

YongYi-coding:
This PR updates `sq_mini_racer` dependency and adds `DS_Store` files to the gitignore. These are things that came up when resolving environment issues.


BintLopez (approved):
🦑

Issue #573, opened 2021-03-01
Pair (MAC + Reprocare): Add color change to balance line when the trigger amount is met


jel535s:
## Context
As a MAC user, I want to see which partner's balances are below the trigger amount as I'm scrolling down my dashboard.

## Process
When a MAC user logs in to the app, they can see all of the partners on their dashboard. 
When a partner's balance drops below the trigger amount, we have functionality that emails MAC's finance team and also creates a red alert on the partner dashboard (see screenshot below). However, when scrolling down through the partner list those alerts are hidden.

To resolve this, we want to add color change to the line showing the partner's name and balance as a MAC user scrolls down their dashboard. The color should be different when the partner's balance drops below the balance alert threshold (yellow?), vs when it's zero or below zero (red?)

## Acceptance Criteria/Definition of Done
When a partner has a balance below the trigger amount, a color change is activated on all MAC user dashboards that highlights the partner's name and their balance (see mockup below). When the balance hits or drops below zero, that color should change again.

## Resources
Red alert on MAC partner dashboard:
![image](https://user-images.githubusercontent.com/53353956/109444232-618bb080-7a02-11eb-8fca-e917d7ca5289.png)

Mockup of new color addition:
![MAC partner dashboard](https://user-images.githubusercontent.com/53353956/109444385-bcbda300-7a02-11eb-9034-c03f01030d12.jpg)

BintLopez:
One question for this task -- should the color be different when below the balance alert threshold vs when it's zero or below?

jel535s:
That's a great question, and I'll adjust it in the scope above!

jel535s:
Nicole merged this PR 6/27

PR #584, opened 2021-03-28
Color change when balance is low

YY-coding:
For issue https://github.com/MidwestAccessCoalition/lockbox_rails/issues/573

@BintLopez There wasn't a balance that's lower than $500 but greater than $0, so I am not sure if the `<% elsif lockbox_partner.low_balance? %>` logic works.
Screen Shot 2021-03-27 at 10 36 59 PM

Probably needs better design too!


YongYi-coding:
here are the color change designs when balance is low or under 0

Screen Shot 2021-03-28 at 3 05 31 PM
Screen Shot 2021-03-28 at 3 05 53 PM
Screen Shot 2021-03-28 at 3 08 02 PM
Screen Shot 2021-03-28 at 3 09 39 PM
Screen Shot 2021-03-28 at 3 09 55 PM

jel535s:
My preference would be to have the text color change as well as the highlighting, as that matches other alerts and styling elsewhere in the app. So, (yellow background + dark yellow text) and (red background + dark red text). Here's an example of a matching alert in the app: 
image

BintLopez:
I agree with Jesse's style preference here!

YongYi-coding:
@BintLopez It looks like this
Screen Shot 2021-04-04 at 11 32 09 PM

app/views/dashboard/_admin_dash.html.erb
```diff
@@ -1,15 +1,26 @@
...
-  <div class="lockbox-activity">
-    <table class="table table-no-border">
+  
+  <% if lockbox_partner.insufficient_funds? %>
+    <h3 class="alert-danger border border-danger px-3">
+  <% elsif lockbox_partner.low_balance? %>
+    <h3 class="alert-warning border border-warning px-3">
+  <% else %>
+    <h3 class="px-3">
+  <% end %>
```
BintLopez, line 14: The logic looks good here! However, this is getting quite a bit of logic mixed into the view, which might make this hard to reason with in the future. Suggestion:  move the logic code into a method in a helper file. For example, you could write something like this in `app/helpers/lockbox_partners_helper.rb`...
```ruby
  def alert_display_header_class(lockbox_partner)
    return unless lockbox_partner.low_balance?
    alert_level = lockbox_partner.insufficient_funds? ? "danger" : "warning"
    "alert-#{alert_level} border border-#{alert_level}"
  end
```
And then in the view you can just do something like... 
```html
<h3 class="<%= alert_display_header_class(lockbox_partner) %> px-3">
```
In Rails, helpers are automatically loaded in view files so that you don't need to require them.

app/views/dashboard/_admin_dash.html.erb
```diff
@@ -1,15 +1,26 @@
...
+      <%= lockbox_partner.name %><span class="float-right">$<%= lockbox_partner.balance %></span>
+    </h3>
+  
+  <div class="lockbox-activity mx-3">
+    <% if lockbox_partner.insufficient_funds? or lockbox_partner.low_balance? %>
```
BintLopez, line 19: You could do something similarly here by moving this to a helper method and then just calling your helper here in the view. Alternatively you could condense the if/else statements into one line like below.

```html
<table class="table table-no-border <%= 'table-top-border' if !lockbox_partner.low_balance? %>">
```

Since we know that if `lockbox_partner.insufficient_funds?` is true, `lockbox_partner.low_balance?` _has_ to be true, we are fine only checking for low_balance here since the behavior is the same.

Also reviewing this PR prompted me to look up the difference between `or` and `||` in Ruby! I typically see `||` more than `or`, and that's because `or` can be tricky -- https://stackoverflow.com/a/2083118. Specifically when used in variable assignment `or` will evaluate to a different value than it will assign to a variable. I played around with this to understand it myself and this was the behavior I saw...
```ruby
def test(a,b)
  a or b
end

test(false,true)
=> true

# It evaluates to true but stores false in the variable
some_variable = false or true
=> true
some_variable
=> false
```

YongYi-coding: @BintLopez Hi Nicole! Thanks for your super helpful and detailed explanation

BintLopez:
The solution looks good to me, we need to get this build passing tests. If you pull from `trunk` branch and then push to this branch, I think the tests will pass.

ALSO, while the solution works and this is in a mergeable state, one thing that could make this PR better is adding tests! Let me know if you want to pair on that :)

BintLopez:
@YY-coding -- Let's get this branch in a passing state so we can merge! I think if you backmerge (while on your branch pull down the most recent trunk changes) and then push to this branch, we should be able to get rid of the failing test problem.

It would be even better if we can add a test to your helper method. What that would entail is creating a file in `spec/helpers` for `spec/helpers/lockbox_partners_helper_spec.rb` and then testing the methods you've added. Either me, @misternu, @kyaroch, or @nsiwnf can answer questions or help with how to write tests in rspec :)

YY-coding:
@BintLopez I just did a back merge

BintLopez:
Going to address the tests in a follow up PR! 🦑

PR #591, opened 2021-04-04
Cherry pick support request form fixes from #574

misternu:
During #520 I discovered problems with the support request form, this PR fixes tooltips and gas reinbursement behavior (on edit, for instance) that are currently broken in prod. Minor visual improvement of lockbox transaction close buttons.


BintLopez:
@misternu I've QA-ed this locally and the solution and behavior looks good. Before we merge, I think this warrants a test so that we know if subsequent code breaks this feature.

Also something I noticed and investigated while QA-ing is that if (for whatever reason) a record exists with a nil distance, the below is the behavior we'd get. This isn't a problem though -- there are only 21 records in the db where distance is nil, all of them are from January, and only one is in pending status. I think it's safe to say that the one pending record won't be in need of edits at the transaction level though. Surfacing this for awareness
image

app/views/lockbox_partners/support_requests/_lockbox_transaction_fields.html.erb
```diff
@@ -1,16 +1,17 @@
-<div class="<%= 'nested-fields row' + (f.object.category == 'gas' ? ' mileage-row' : '') %>">
+<% gas_select_name = LockboxTransaction::GAS_SELECT_NAME %>
+<div class="<%= 'nested-fields row' + (f.object.category == gas_select_name ? ' mileage-row' : '') %>">
```
scottyschup, line 2: Minor style/readability comment: Rails 6 added a handy [`class_names` helper](https://stackoverflow.com/a/63984432/3251463) to make conditional uses of CSS classes a little more readable:
```suggestion
<div class=<%= class_names('nested-fields', 'row', { 'mileage-row' => f.object.category == gas_select_name }) %>>
```

Alternately, I find string interpolation a little easier to read than concatenation.
```suggestion
<div class=<%= "nested-fields row #{' mileage-row' if f.object.category == gas_select_name}") %>>
```

But at the very least, the double-quotes around the class value are unnecessary and the ternary can just be `' mileage-row' if f.object.category == gas_select_name` to avoid the `: ""` bit.

scottyschup (approved):
🦉 🎉 
Just one non-blocking style/readability comment.

misternu:
@BintLopez gtg pending re-approval

BintLopez (approved):
🦑

Issue #589, opened 2021-03-28
Dev: Combine "Edit Profile" and "Account Settings"/"Update Password"


jel535s:
## Context
Currently in the app, users have 2 places where they can make updates unique to their profile.
"Edit Profile" takes them to a page headed "Update User", where they can change their profile name and time zone (see screenshot in Resources below). Soon this page will also include notification preferences for Funder users only, once work on #587 is complete.
"Update Password" takes them to a page headed "Update Password" where they can.... change their password. (screenshot below in Resources)

We want to combine these two pages into one page called "Account Settings".

## Process
We want to combine these two pages into one page called "Account Settings". New designs shouldn't be needed since the page designs should not change much from how they look currently. The order on the page should be:

1. Name field
2. timezone field
3. password fields
4. if a Funder user, notification preferences (if you're starting this before #587 is completed, give the PM and the dev on that ticket a heads up as it will affect their work).

## Acceptance Criteria/Definition of Done
On this page, every user should be able to:

- change their name
- change their timezone
- update their password

And once #587 is complete, on this page Funder users should also be able to:

- choose between two email notification options

## Resources
Edit Profile page:
![image](https://user-images.githubusercontent.com/53353956/112771804-3d1bf780-8ff3-11eb-820b-e758f93a28a1.png)
Update Password page:
![image](https://user-images.githubusercontent.com/53353956/112771879-97b55380-8ff3-11eb-82e9-7aa6260b582b.png)

nsiwnf:
@jel535s Should "Account settings" be the where the Edit Profile button is or the Update Password  button?

jel535s:
@nsiwnf let's put it where the "Update Password" button is- on the right hand side. And then the only thing under their contact info will be "Edit Contact Information".

Issue #586, opened 2021-03-28
Backend Dev: Funder users can configure their own email alerts


jel535s:
## Context
Currently, all Support Request actions trigger an email to all Funder users from a certain funder, regardless of who created the Support Request. We had a user provide direct feedback that they would like to be able to configure this to only see Support Requests they created.

The design here is still pending but the backend work can start now! The frontend work will be handled on TICKET and you'll want to pair to hook up the frontend dev work with the backend work.

## Process
Based on the designs from #585, on the Edit Profile page, add a new field that will allow Funder users to adjust their email notifications.

Potential wording-
Email Notifications Preference:
I want emails about every Support Request
or
I only want emails about Support Requests I created.

## Acceptance Criteria/Definition of Done
Funder users (currently just MAC users) are able to configure which email alerts they get, by using the new field in their Edit Profile page. Effective as soon as their preference is saved, the email alerts are updated to only send what they request.

## Resources
Pending from #585

misternu:
@jel535s @jasonwynkoop86 Currently reviewing what's done of #589, and poking around these other changes and discussions. We do not currently store the address of a user, and I'm not sure there's a compelling reason to do so.

It would also be useful for devs to know where/if a user is redirected after editing the password or user information. Redirects to the main page made sense when we were editing less information, and this may no longer make sense.

If saving/submitting email notifications is going to be attached to the user info submit button, are we sure that the user will know that their email preference hasn't been saved?

jasonwynkoop86:
When things are updated currently it returns to the main page with a green attention box. 
![image](https://user-images.githubusercontent.com/82782739/120871339-f3a2d800-c560-11eb-80a0-58c02e20e8cc.png)
Can we refresh the screen on update to show an attention box like the image above for each parameter that has been updated?
And another update for updating time zone, name, and email preferences. 

We don't need to have an address or change address form, but some of the partners have addresses tied to the accounts and if they were to be edited it should be through this universal place. Keeping all profile info that is editable in one place.

nsiwnf:
@jel535s @BintLopez  Admins/MAC users can select to either get notifications from all user actions or just other admin actions?
We have note notification logic as 
>MAC users only get emailed about manual notes
MAC users don't get emailed about notes they wrote
Partner users get emailed about all notes

And insufficient funds/low balance as 
>All admins and lockbox email

BintLopez:
[...]

Issue #585, opened 2021-03-28
Design: Funder users can configure their own email alerts


jel535s:
## Context
Currently, all Support Request actions trigger an email to all Funder users from a certain funder, regardless of who created the Support Request. We had a user provide direct feedback that they would like to be able to configure this to only see Support Requests they created.

## Process
On the Edit Profile page, design a new field that will allow Funder users to adjust their email notifications.

Potential wording-
Email Notifications Preference:
I want emails about every Support Request
or 
I only want emails about Support Requests I created.

## Acceptance Criteria/Definition of Done
We have a design mockup that the devs can use to model their work.

## Resources
Current Edit Profile screen in the app:
![image](https://user-images.githubusercontent.com/53353956/112770184-1b1e7700-8feb-11eb-84fa-a630df2ce7d4.png)

jasonwynkoop86:
i think we should combine this ticket with #599
since we are editing the account screen in both tickets. 
added a drop down menu for selecting notifications settings for support requests. 
added 2 menu options. 
Receive All Support Requests Email Notifications
Only Receive Support Request Email Notifications I Created

![image](https://user-images.githubusercontent.com/82782739/119278702-93b94280-bbec-11eb-98c6-3782394ca7d2.png)

saving the option is the same as saving any edits to your account. 
![image](https://user-images.githubusercontent.com/82782739/119278720-b2b7d480-bbec-11eb-86d8-ebaa68c74c32.png)

jel535s:
These designs look great to me! I also noted on the #599 that they may want to reference these designs or even combine the work done there with #586.

PR #599, opened 2021-05-16
combine profile and password

YY-coding:
## Changelog
- I changed the update password heading to account setting
- I put the logic from edit profile to account setting where has the update password logic

## Link to issue:  
https://github.com/MidwestAccessCoalition/lockbox_rails/issues/589

- N/A

## Relevant Screenshots: 
When you clicked setting, you will see this
Screen Shot 2021-05-16 at 6 41 15 PM

when you click account setting, you will see this
Screen Shot 2021-05-16 at 6 40 34 PM

## Are you ready for review?:

- [ ] Added relevant tests
- [x] Linked PR to the issue

### DIFF
--- a/app/views/devise/registrations/edit.html.erb
+++ b/app/views/devise/registrations/edit.html.erb
@@ -1,5 +1,6 @@
 <div class="container">
   <%= form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :put, class: 'form' }) do |f| %> 
+    <%= render "devise/shared/error_messages", resource: resource %>
     <div>
       <h2>Update Password</h2>
     </div>
@@ -33,19 +34,6 @@
   <% end %>
 
   <%= form_for @user, html: { class: 'form' } do |f| %>
-    <% if @user.errors.any? %>
-      <div id="error_explanation">
-        <h2>
-          <%= @user.errors.count %> error(s) prohibited this user from being saved:
-        </h2>
-        <ul>
-          <% @user.errors.full_messages.each do |error_message| %>
-            <li><%= error_message %></li>
-          <% end %>
-        </ul>
-      </div>
-    <% end %>
-
     <h2>Update User</h2>
     <div class="form-group">
       <%= f.label :name %>
--- a/app/views/devise/registrations/edit.html.erb
+++ b/app/views/devise/registrations/edit.html.erb
@@ -1,36 +1,71 @@
-<%= form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :put, class: 'form' }) do |f| %>
-  <%= render "devise/shared/error_messages", resource: resource %>
+<div class="container">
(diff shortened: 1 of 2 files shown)


jasonwynkoop86:
Here is my take on the account screen. 
Changes I made. 
Delete edit password button on the right nav. 

Change edit profile to edit account settings. 

Combine account info screens for Change password, name, gmt, and address info.

image

jel535s:
@YY-coding FYI: @jasonwynkoop86 added some designs to #585 that might affect your work here. He also noted that it might make sense to combine this PR with the work spec'd out on #586 . Let me know what you think- is it easier to build that request in to this PR, or would you rather proceed with the work you've already started here and then pick up #586 after?

misternu:
Screen Shot 2021-05-31 at 9 31 56 AM
I found an issue where if there is an error it is displayed on both forms. Theres some other stuff we should try to work out about upcoming work for this page that is currently being designed/redesigned. Will mention during stand and we can probably pair on it and talk to the appropriate people

BintLopez:
Discussed issue with duplicate error messages and decided to simplify by only having one error message div at the top of the page. Note -- fields with error should still have a red border

BintLopez:
Also discussed the ability to change an address. Since the address is on the LockboxPartner (soon to be renamed Organization) instead of on the User, it should not be editable on this page. cc @jasonwynkoop86

YY-coding:
I can only put the error message in the middle. If I put the message on the top, after submitting the change, this error appears
Screen Shot 2021-06-06 at 7 01 50 PM

with the error message in the middle, after refreshing the page, the following error appears
Screen Shot 2021-06-06 at 7 51 14 PM

This is how the page looks like form is not validated
Screen Shot 2021-06-06 at 7 00 19 PM

app/views/devise/registrations/edit.html.erb
```diff
@@ -1,36 +1,69 @@
...
+        <h2>
+          <%= @user.errors.count %> error(s) prohibited this user from being saved:
+        </h2>
+        <ul>
+          <% @user.errors.full_messages.each do |error_message| %>
+            <li><%= error_message %></li>
+          <% end %>
+        </ul>
+      </div>
+    <% end %>
```
BintLopez, line 47: I think instead what we can do is delete lines 36-47 and re-insert `<%= render "devise/shared/error_messages", resource: resource %>` between lines 2 and 3.

YY-coding: @BintLopez I just did what you suggested. The form validation message is on the top of the page now as you can see in the image below. 
Screen Shot 2021-06-13 at 1 17 17 PM

But if I click the refresh arrow while the validation message is on page, it gives me a route error as you can see on the image below

Screen Shot 2021-06-13 at 1 17 40 PM

I also try this in the trunk, it has the same issues. I hope it makes sense.

misternu:
The described issue is a part of normal rails/devise behavior and is in current trunk, merging this as is with a possible upcoming cleanup PR

misternu (approved):
🦉

PR #616, opened 2021-07-18
Nav styles 610

YY-coding:

## Changelog
- change breakpoint for the #navbar-drawer.container
- .navbar-subcontainer width is 100%
- when max-width is 767.9px, all buttons' width is 100% of their parent .navbar-subcontainer (note that I made the max-width at 767.9px because I would like the navbar style stays the same in ipad screen size which is 768px. Let me know if you don't think that's a good idea)

## Link to issue:  
[Fixes issue #N/A](https://github.com/MidwestAccessCoalition/lockbox_rails/issues/610)

- N/A

## Relevant Screenshots: 
- Funder page
Screen Shot 2021-07-18 at 11 18 12 AM

- Admin page
Screen Shot 2021-07-18 at 11 20 29 AM

## Are you ready for review?:

- [ ] Added relevant tests
- [x] Linked PR to the issue


BintLopez:
I think this looks great! Two things...

1. For the lockbox partner's nav, I'm wondering if it makes sense to switch the order of the buttons so that logout is the one on the bottom. @jel535s @gujacq @jasonwynkoop86 -- thoughts?
2. While we're here can we change the button text of "Edit Account setting" to "Edit Account Settings"
image

jel535s:
I agree on both points!

Issue #577, opened 2021-03-07
UI/Frontend: allow an “admin” user to delete a reconciliation action through the UI


jel535s:
## Context
Blocked by designs from #578 
We’ve had multiple occasions now where a clinic partner inputs the incorrect amount for reconciliation. The best resolution for this is to just delete the event, which will just trigger reconciliation alerts for that partner again the next time they log in. Currently the only way we’re resolving this is super manually through heroku console, which is fairly cumbersome for the team.

jel535s:
Designs are ready! https://www.figma.com/proto/xhSGqoMxLZOqQX5d3eS7DF/LockBox-Reconciliation?node-id=2%3A2&scaling=min-zoom&page-id=0%3A1

PR #618, opened 2021-07-18
577 allow admin delete reconciliation

emmajuettner:
(WIP currently, need to check on some tests that are failing after these changes)

## Changelog
- Added "Delete Reconciliation" button in "Actions" column of activity list, which goes to "Delete Reconciliation" page
- Canceled reconciliations show up as crossed out in the activity list (amount is not included in total balance, but it does show up in the table)
- Canceled reconciliations have "Deleted" icon in "Actions" column, which has a tooltip that says when/who deleted it (this is visible to both admins and partners)

## Link to issue:  
Fixes issue #577 

- Reconcile as a lockbox partner
- Log in as admin user and delete the reconciliation

## Relevant Screenshots: 
A reconciliation before deleting, with admin user logged in:
image
Delete Reconciliation page:
image
Reconciliation after deleting:
image

## Are you ready for review?:

- [ ] Added relevant tests
- [ ] Linked PR to the issue

### DIFF
--- a/app/assets/stylesheets/lockbox_activity.scss
+++ b/app/assets/stylesheets/lockbox_activity.scss
@@ -17,6 +17,14 @@ tr.canceled {
   font-style: italic;
 }
 
+.deleted-reconciliation-tooltip-icon {
+  color: $color-primary;
+}
+
+.deleted-reconciliation-tooltip-text {
+  color: $color-primary;
+}
+
 .action-label {
   display: none;
 }
--- a/app/controllers/lockbox_actions_controller.rb
+++ b/app/controllers/lockbox_actions_controller.rb
@@ -19,7 +19,7 @@ class LockboxActionsController < ApplicationController
   end
 
   def update_params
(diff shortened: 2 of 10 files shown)


emmajuettner:
This PR is ready for review! One thing in particular that I would like some feedback on is the routing for the delete reconciliation page... what I have written works just fine when I test it, but it seems like there must be a cleaner way to do the same thing. Same goes for the strikethroughs on canceled reconciliations-- I feel like there's got to be a way of accomplishing this with less cluttered-looking code, but I'm not sure what that is. (I'm sure there's other things I will need to fix/improve as well, but these are just the main things that stand out to me.)

One other note-- the design for this has a button to add a note to a deleted reconciliation, but this would require some fairly substantial changes as currently notes are not something that can be associated with a reconciliation. I brought it up a while ago in one of the meetings and we decided that the ability to add notes to reconciliations was not really needed, so just wanted to document that here.

PR #623, opened 2021-08-08
Automatically round up individual transactions that a user edits

misternu:
## Changelog
- Round each individual transaction that a user edits up to the nearest dollar
- Round mileage reinbursement transactions to the nearest dollar when edited
- DO NOT round up transactions that already have cents and are not edited (eg, user edits other parts of support request, does not touch transactions, and saves. Transactions are not changed)

## Link to issue: 
https://github.com/MidwestAccessCoalition/lockbox_rails/issues/520


misternu:
I'm going to try to write a document explaining this problem and its current solution, to be linked here

